PR #169, opened 2014-09-25
Adding syslog support.

brookerj11211:
Add support for sending logs directly to syslog using
the python logging libraries "SysLogHander"

The change introduces a "logtype" feature whereby the user can
choose to log to a file, syslog or stderr

The existing functionality is retained, in as much as simply
specifying a logfile will cause it to act as if you also specified
the "file" logtype.  Also if "file" logtype is specified but no file is
specified it will default to logging to stderr.

### DIFF
--- a/curator/curator_script.py
+++ b/curator/curator_script.py
@@ -4,6 +4,7 @@ import os
 import sys
 import time
 import logging
+import logging.handlers
 from datetime import timedelta, datetime, date
 
 import elasticsearch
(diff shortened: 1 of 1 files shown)


magnusbaeck:
We'd obviously have to catch the exception and perhaps truncate the message before retrying (not sure how we'd do that though – subclass SysLogHandler?), but if that only happens for debug-level messages at least I'd be fine with that. If you need debug-level logging you're probably not interested in logging to syslog anyway – it's typically things you want in your terminal or in a temporary local logfile. IMHO it's a reasonable limitation if properly documented.

untergeek:
Closing this as v3 has a completely changed code base.  If you'd like to try this in v3, please submit a new PR.

Issue #246, opened 2015-01-11
Skip replica count change if not needed


bobrik:
I have 400+ daily indices and curator tries to set new replica count for every index that matched desired time span. This takes a while and could be avoided by simple check for current replica count.

```
2015-01-11 09:34:40,932 INFO      change_replicas operation succeeded on statistics-20131004
...
2015-01-11 09:39:50,985 INFO      change_replicas operation succeeded on statistics-20141004
```

5+ minutes to do nothing.

I can send a PR if you're interested.

untergeek:
@bobrik This will be easily addressed with the pipelining feature pending in 3.0.

untergeek:
However, if you want to do a short-term PR for count validation, I will accept it and merge it to master.

untergeek:
While checking replica count could help, most of the need for this is obviated with better index filtering, in v3.  Also, all matching indices get the replica count updated simultaneously in v3.

I have removed these checks for now.  If they are desired, we can re-add them to v3.

PR #247, opened 2015-01-13
not updating replicase if not needed

bobrik:
This addresses #246.


bobrik:
Log message updated.

untergeek:
LGTM!

Issue #248, opened 2015-01-13
Logstash log format should always produce proper JSON


magnusbaeck:
With `--logformat logstash`, Curator uses the standard log formatter ([logging.Formatter](https://docs.python.org/2/library/logging.html#logging.Formatter)) and just configures the format string to produces something that's _usually_ valid JSON. However, if the message or another field from the log record contains e.g. a backslash or a double quote this results in broken JSON. We should instead use a formatter that always produces proper JSON (e.g. using [Python's json module](https://docs.python.org/2/library/json.html)).

I'm almost done with a patch for this. Does the Elasticsearch CLA cover Curator as well?

untergeek:
That is why the code was updated with `json.dumps` in the places where that could happen.

https://github.com/elasticsearch/curator/search?utf8=✓&q=json.dumps

It's one of the changes in [Curator 2.1.0](https://github.com/elasticsearch/curator/blob/059b99e3bafc3d86288af059be748455be03533c/Changelog.rst#210-30-december-2014)

If there are other places this can happen that I have missed, I would prefer to keep the usage similar.

magnusbaeck:
Don't know how I managed to miss that bug. I thought I searched for "logstash".

Wait, what? You prefer having to manually insert json.dumps() calls for log messages that you think can contain characters that screw up the output?

Maybe I'm too tired, but I don't see how it's supposed to work – json.dumps() obviously returns a double quoted, but the formatter's format string also includes literal double quotes for the message, resulting in double double quotes in the final rendered message. Here's an example message from a stock Curator 2.1.1 with `--loglevel DEBUG` and `--logformat logstash`, and it's not valid JSON:

    {"@timestamp":"2015-01-13T23:35:10.623Z", "loglevel":"DEBUG", "name":"curator", "function":"filter_by_timestamp", "linenum":"375", "message":""Unable to match logstash-2014.07.31_ with regular expression ^logstash-(\\d{4}\\.\\d{2}\\.\\d{2})$.  Error: 'NoneType' object has no attribute 'group'""}

untergeek:
Looks like `json.dumps` didn't wrap that properly.  I wonder why not. Looks like a better solution is in order.

So, what does your PR look like?

magnusbaeck:
[...]

PR #250, opened 2015-01-14
Produce valid JSON for all input with the logstash log format.

magnusbaeck:
The old method of constructing the log messages for the logstash
format produced invalid JSON for certain input. We avoid this by
instead using Python's standard json library for producing the JSON
string. As a bonus this allows us to drop several json.dumps() calls
in the log statements themselves. This addresses issue #248.

### DIFF
--- a/curator/curator.py
+++ b/curator/curator.py
@@ -1,7 +1,6 @@
 import time
 import logging
 import re
-import json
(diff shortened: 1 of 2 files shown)


untergeek:
lgtm!

PR #258, opened 2015-01-23
Add --not-older-than argument

Dinoshauer:
This PR adds a ``--not-older-than`` argument for the ``allocation`` sub-command, useful for when you want to work on a sub-set of indices.
Our use case is a more-than-2-tiered cluster and we want to move indices between our 3 tiers and ``--not-older-than`` will ease that work.

I've only added the option to the ``allocation`` sub-command because that is our use-case, implementing the same functionality for the other sub-commands should be rather easy.

I've added a unittest folder as well as I didn't see a reason to do a full integration test on this small piece of logic.

### DIFF
--- a/curator/curator.py
+++ b/curator/curator.py
@@ -328,8 +328,9 @@ def get_object_list(client, data_type='index', prefix='logstash-', suffix='', re
 # Filtering
 ## By timestamp
 def filter_by_timestamp(object_list=[], timestring=None, time_unit='days',
-                        older_than=999999, prefix='logstash-', suffix='',
-                        snapshot_prefix='curator-', utc_now=None, **kwargs):
+                        older_than=None, not_older_than=None,
+                        prefix='logstash-', suffix='', utc_now=None,
+                        snapshot_prefix='curator-', **kwargs):
     """
     Pass in a list of indices or snapshots. Return a list of objects older
     than *n* ``time_unit``\s matching ``prefix``, ``timestring``, and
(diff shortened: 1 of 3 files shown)


untergeek:
I appreciate this!  Thank you for making Curator more awesome.

Unfortunately, I am changing the whole pipeline for Curator 3.0 and this functionality will be included in a different way.  As such, I will not be merging this into master right now because it will be in conflict with the future release.  By all means, keep using your forked branch to have this functionality until then.

The new 3.0 pipeline will allow multiple filtration methods simultaneously.  For example,  `--older-than` and `--newer-than` could both be used at once to have a custom filtered list of indices.

Dinoshauer:
Ah! Well, that's awesome!

Do you have any ETA on the release of 3.0?

untergeek:
I'm spending most of my time on Logstash 1.5 with the rest of my team.  I wish I could give you more information, but it will be a month or two, I fear.

Dinoshauer:
No worries, I am no stranger to prioritizing :) I'll keep a look out for 3.0 and keep using this branch until then! You can close off this PR if you wish.

untergeek:
v3 is at RC1 right now.  Please see if the combination of `--older-than` and `--newer-than` work for you now, along with the other pipelined filters available.

Dinoshauer:
Just tried it out and it worked awesome :+1: - Can I subscribe to something somewhere to be alerted of when this goes into production?

PR #261, opened 2015-01-29
Avoid removal of kibana index for kibane 4

malagoli:


untergeek:
Thanks for adding this.

PR #262, opened 2015-01-30
Add functionality to open indices and set shards per node

rtkmhart:
Add functionality to open indices to compliment the close option.

Add functionality to set the number of shards per node for an index. We require this to rebalance our cluster fairly regularly, and hopefully it's useful to others as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ See the [Curator Wiki](http://github.com/elasticsearch/curator/wiki) on Github f
 * bloom - Disable bloom filter cache for expired indices
 * close - Close indices
 * delete - Delete indices
+* open - Open indices
 * optimize - Optimize (Lucene forceMerge) indices
 * show - Show indices or snapshots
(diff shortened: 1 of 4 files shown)


untergeek:
Thanks for the addition, and for making curator more awesome.

I'm about to totally change-up how curator works for 3.0.  I'm happy to merge your changes to `curator.py` as the API can use these methods. I have twice denied adding an "open" command in the past as it is not typical for people to do this frequently. But I am going to want it for 3.0, so thanks for your addition.

However, anything you add to `curator_script.py` will likely be obliterated.  I recommend keeping your own fork of `curator_script.py` for the time being and removing it from this PR.

rtkmhart:
Ok, since I have an urgent need for this I'll do something else, but leave this PR for you to cherry pick from.

untergeek:
As "open" is now in v3, I'm going to close this with an invitation to do another pull request for the shards-per-node functionality against the v3 code base.

PR #283, opened 2015-03-05
Allow allocation and replicas operations for closed indices

### LINKED ISSUE #279: Optionally allow some changes to closed indices
I think there are some cases when it would actually be useful to be able to apply some of the changes curator supports even to closed indices.

For example when someone uses an age-based data retention plan, like having indexer and storage nodes, and keeping old indices around in a closed state for a while before they actually get deleted/backed up. I believe that is quite a common setup when dealing with logs or other time-based data, especially on a high volume.

In the above scenario it can be the case that the allocation rules are changing between the closing and the reopening of old indices, so they might start recovering to the wrong nodes (which can be painful if huge amounts of data is involved).

This can be prevented if there would be an `--even-for-closed-indices` switch for e.g. the `allocation` command, but it would possibly be useful for some other operations as well _as an option_ (`replicas`? maybe `alias`? ). So interim allocation rule changes are constantly could be applied even to closed indices if that makes sense for someone's use case.

I have a simple implementation of this at hand against v2, and I'm willing to convert it to v3 if it sounds like a good idea.

ferki:
This is an attempt to fix #279. @untergeek: could you please review the changes? Especially the test for changing the replica count of a closed index, as it seems to be already possible without any code change despite calling `prune_closed` (but maybe it's just late and it's not a bug :) So I omitted changing this [`prune_closed`](https://github.com/elasticsearch/curator/blob/master/curator/api/replicas.py#L20) to `ensure_list` for now, but I agree that it would be proper to change that.

ps.: I also think that adding a `prune_replicated` function would be useful to skip the indices that already has the requested replica count (as it is slow to forcibly apply it). I can send a separate PR for that, but since its implementation would probably be related to this changeset, I guess it's a good idea to get this done first.

### DIFF
--- a/test/integration/test_api_commands.py
+++ b/test/integration/test_api_commands.py
@@ -102,6 +102,31 @@ class TestChangeReplicas(CuratorTestCase):
 
         self.assertEquals('0', self.client.indices.get_settings(index='test_index')['test_index']['settings']['index']['number_of_replicas'])
 
+    def test_closed_index_replicas_can_be_modified(self):
+        self.create_index('test_index')
+
+        self.client.indices.close(index='test_index')
+        index_metadata = self.client.cluster.state(
+            index='test_index',
+            metric='metadata',
+        )
+        self.assertEquals('close', index_metadata['metadata']['indices']['test_index']['state'])
+
+        self.assertEquals('0', self.client.indices.get_settings(index='test_index')['test_index']['settings']['index']['number_of_replicas'])
+
+        curator.change_replicas(self.client, 'test_index', replicas=1)
+
+        self.assertEquals('1', self.client.indices.get_settings(index='test_index')['test_index']['settings']['index']['number_of_replicas'])
+
+        self.client.indices.open(index='test_index')
+        index_metadata = self.client.cluster.state(
+            index='test_index',
+            metric='metadata',
+        )
+        self.assertEquals('open', index_metadata['metadata']['indices']['test_index']['state'])
+
+        self.assertEquals('1', self.client.indices.get_settings(index='test_index')['test_index']['settings']['index']['number_of_replicas'])
+ 
 class TestChangeAllocation(CuratorTestCase):
     def test_index_allocation_can_be_modified(self):
         self.create_index('test_index')
--- a/test/integration/test_api_commands.py
+++ b/test/integration/test_api_commands.py
@@ -102,6 +102,37 @@ class TestChangeReplicas(CuratorTestCase):
 
         self.assertEquals('0', self.client.indices.get_settings(index='test_index')['test_index']['settings']['index']['number_of_replicas'])
 
+class TestChangeAllocation(CuratorTestCase):
+    def test_index_allocation_can_be_modified(self):
+        self.create_index('test_index')
+
+        curator.apply_allocation_rule(self.client, 'test_index', rule="key=value")
+
+        self.assertEquals('value', self.client.indices.get_settings(index='test_index')['test_index']['settings']['index']['routing']['allocation']['require']['key'])
+
+    def test_closed_index_allocation_can_be_modified(self):
+        self.create_index('test_index')
+
+        self.client.indices.close(index='test_index')
+        index_metadata = self.client.cluster.state(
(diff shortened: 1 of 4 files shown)


untergeek:
Please add the change to `ensure_list` to https://github.com/elasticsearch/curator/blob/master/curator/api/replicas.py#L20 as well.

I don't think a method to skip indices with the correct number of replicas is a bad idea per se, but I don't think it's as necessary now that the replica count is applied to all indices at once.  In the past it iterated the indices one at a time.  I'm not 100%, but I don't think anything untoward will happen telling an index with 0 replicas to have 0 again.  In any case, the settings are applied simultaneously now for all but alias and optimize operations.  It will also iterate with bloom filter disabling, but only if you specify a `--delay` argument.  But since bloom filters are disabled since 1.4, this is a non-issue for most.

test/integration/test_api_commands.py
```diff
@@ -102,6 +102,44 @@ def test_index_replicas_untouched(self):
...
+        self.client.indices.close(index='test_index')
+        index_metadata = self.client.cluster.state(
+            index='test_index',
+            metric='metadata',
+        )
+        self.assertEquals('close', index_metadata['metadata']['indices']['test_index']['state'])
+
+        curator.apply_allocation_rule(self.client, 'test_index', rule="key=value")
+
+        self.assertEquals('value', self.client.indices.get_settings(index='test_index')['test_index']['settings']['index']['routing']['allocation']['require']['key'])
```
untergeek: I think this test should open the index after applying the rule also, and then test to ensure the setting is present after opening.

test/integration/test_api_commands.py
```diff
@@ -102,6 +102,44 @@ def test_index_replicas_untouched(self):
...
+            index='test_index',
+            metric='metadata',
+        )
+        self.assertEquals('close', index_metadata['metadata']['indices']['test_index']['state'])
+
+        self.assertEquals('0', self.client.indices.get_settings(index='test_index')['test_index']['settings']['index']['number_of_replicas'])
+
+        curator.change_replicas(self.client, 'test_index', replicas=1)
+
+        self.assertEquals('1', self.client.indices.get_settings(index='test_index')['test_index']['settings']['index']['number_of_replicas'])
```
untergeek: I think this test should open the index after changing the replica count and ensure the setting persists after re-opening.

ferki:
Thanks for the feedback! I pushed the changes you requested.

And yes, with the one-step replica count manipulation it's not necessary to skip the indices that already has the requested replica count.

ps.: are there something wrong with our CCLA or does it just need some time before tests start passing?

untergeek:
I'll have to ask about the CCLA.  What's your organization?

Are the tests all passing for you now?  I'll download and test myself if necessary.

Also, are you fortunate enough to have both a 2.7 and 3.x python instance installed?  Our CI tests 2.7 and 3.3.  I test on my own with 2.7.2 and 3.4.2.

ferki:
The company name on the CCLA is `adjust GmbH`.

Tests are passing for my local single-node elasticsearch-1.4.4 with both python-2.7.9 and 3.3.5.

untergeek:
Hmmm.  I'm trying to manually verify. Our manual verification tool prompts me:

> Type user's name, e-mail or Github nickname and press Enter...

I've entered `ferki`, `@ferki`, `Ferenc Erki`, and `adjust GmbH` and none of those are working.  Was there a different email entered that I can try?

ferki:
On the last page of the CCLA, I gave my name (Ferenc Erki), my email (ferki@adjust.com - I also use this for the commits I send), and my github username (ferki). I used the same details at the start of the document as contact info. If that helps investigating, I can also send you other data from the signed document I received via email (like transaction ID, etc.) - just let me know if I can help in any way with this.

untergeek:
Strange.  Please send the transaction ID and anything else to my email, aaron.mildenstein at `elasticsearch.com`

We'll get to the bottom of this.

untergeek:
Manually confirmed for now.  Will investigate why the CCLA isn't being recognized later.

PR #337, opened 2015-04-02
fix(cli.es_repo_mgr): Ensure proper parent CLI parameters are passed int...

### LINKED ISSUE #336: es_repo_mgr cli is ignoring parent options for create sub-command
Either I'm doing something silly, or `click`/`es_repo_mgr` is not passing parent command command-line options to the `create` sub-command.

I was able to validate this by simply putting `print('connecting...', kwargs)` on [utils.py#L105](https://github.com/elastic/curator/blob/89e84f960b0188e6dd38dd48d66a4a67a4bb66dc/curator/cli/utils.py#L105).

Success with `show` command:
```
% es_repo_mgr --host foobar show                                                                                                                                             
('connecting...', {'url_prefix': u'', 'http_auth': None, 'loglevel': u'INFO', 'host': u'foobar', 'logformat': u'default', 'timeout': 30, 'debug': False, 'use_ssl': False, 'logfile': None, 'port': 9200})
```

Error with `create` command:

```
% es_repo_mgr --host foobar create fs --repository foo --location /tmp/foo
('connecting...', {})
```

I'm not familiar enough with the `click` library to tell if maybe someone just misconfigured it, but I know enough about python to be dangerous. Let me know if you need more info, or if I can help otherwise.

feltnerm:
...o create sub-command

fixes #336

### DIFF
--- a/test/integration/test_cli_commands.py
+++ b/test/integration/test_cli_commands.py
@@ -887,7 +887,7 @@ class TestCLIRepositoryCreate(CuratorTestCase):
                         'create',
                         'fs',
                         '--repository', self.args['repository'],
-                        '--location', '/tmp'
+                        '--location', self.args['location']
                     ],
                     obj={"filters":[]})
         self.assertTrue(1, len(self.client.snapshot.get_repository(repository=self.args['repository'])))
--- a/test/integration/test_cli_commands.py
+++ b/test/integration/test_cli_commands.py
@@ -874,3 +874,21 @@ class TestCLILogging(CuratorTestCase):
                     obj={"filters":[]})
         output = sorted(result.output.splitlines(), reverse=True)[:4]
         self.assertEqual(expected, output)
+
+class TestCLIRepositoryCreate(CuratorTestCase):
+    def test_create_fs_repository(self):
+        test = clicktest.CliRunner()
+        result = test.invoke(
+                    curator.repomgrcli,
+                    [
+                        '--debug',
+                        '--logfile', os.devnull,
+                        '--host', host,
+                        '--port', str(port),
+                        'create',
+                        'fs',
+                        '--repository', self.args['repository'],
+                        '--location', '/tmp'
+                    ],
+                    obj={"filters":[]})
+        self.assertTrue(1, len(self.client.snapshot.get_repository(repository=self.args['repository'])))
--- a/test/integration/test_cli_commands.py
+++ b/test/integration/test_cli_commands.py
@@ -881,7 +881,6 @@ class TestCLIRepositoryCreate(CuratorTestCase):
         result = test.invoke(
                     curator.repomgrcli,
                     [
-                        '--debug',
                         '--logfile', os.devnull,
                         '--host', host,
                         '--port', str(port),
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -58,3 +58,4 @@ Contributors:
 * Michael-Keith Bernard (SegFaultAX)
 * Simon Lundström (simmel)
 * (pkr1234)
+* Mark Feltner (feltnerm)
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -10,7 +10,8 @@ Changelog
 
  * If an index is closed, indicate in ``show`` and ``--dry-run`` output.
    Reported in #327. (untergeek)
- * Fix issue where CLI parameters were not being passed to sub-commands.
+ * Fix issue where CLI parameters were not being passed to the ``es_repo_mgr``
+   create sub-command.
    Reported in #337. (feltnerm)
 
 3.0.3 (27 Mar 2015)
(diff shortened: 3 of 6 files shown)


feltnerm:
Just signed CLA.

untergeek:
An integration test or two for this should be added (and I apologize, that's probably why this happened at all--this is the one piece of code not tested).  I can help out with this if you like.

Any integration test would belong in https://github.com/elastic/curator/blob/master/test/integration/test_cli_commands.py

The other classes there show how to mimic command-line options in testing.  You'll want to create a repository, then test for its existence (I know, redundant since it's validated, but you can re-use the same call in the test).

Feel free to ask any questions.

feltnerm:
Not my greatest code, but it's a start. 

I was/am a bit confused about how to test changing the host parameter (for the CLI args). It appears that the integration tests run against a locally running elasticsearch instance (is that true?). It'd be nice to mock this out somehow so we could see if using different hosts had effect.

test/integration/test_cli_commands.py
```diff
@@ -874,3 +874,21 @@ def test_logging_with_logformat_logstash_flag(self):
...
         output = sorted(result.output.splitlines(), reverse=True)[:4]
         self.assertEqual(expected, output)
+
+class TestCLIRepositoryCreate(CuratorTestCase):
+    def test_create_fs_repository(self):
+        test = clicktest.CliRunner()
+        result = test.invoke(
+                    curator.repomgrcli,
+                    [
+                        '--debug',
```
untergeek: This line is probably unnecessary, since we're feeding the logs to /dev/null

feltnerm: D'oh, that was for my eyes only.

test/integration/test_cli_commands.py
```diff
@@ -874,3 +874,20 @@ def test_logging_with_logformat_logstash_flag(self):
...
+        result = test.invoke(
+                    curator.repomgrcli,
+                    [
+                        '--logfile', os.devnull,
+                        '--host', host,
+                        '--port', str(port),
+                        'create',
+                        'fs',
+                        '--repository', self.args['repository'],
+                        '--location', '/tmp'
```
feltnerm: I had to hard-code this as '/tmp' because I am running elasticsearch in a virtual machine and happen to have my ports forwarded (so localhost:9200 is hitting ES on my VM). The test utilities attempt to create a temporary folder, but that folder only exists on my host machine. '/tmp' seemed reasonable, for now.

test/integration/test_cli_commands.py
```diff
@@ -874,3 +874,20 @@ def test_logging_with_logformat_logstash_flag(self):
...
         self.assertEqual(expected, output)
+
+class TestCLIRepositoryCreate(CuratorTestCase):
+    def test_create_fs_repository(self):
+        test = clicktest.CliRunner()
+        result = test.invoke(
+                    curator.repomgrcli,
+                    [
+                        '--logfile', os.devnull,
+                        '--host', host,
```
feltnerm: Testing this against `localhost` does not really tell if the erroneous behavior I experience is fixed. It'd be nice to be able to test against _any_ host, even if the connection fails.

test/integration/test_cli_commands.py
```diff
@@ -874,3 +874,21 @@ def test_logging_with_logformat_logstash_flag(self):
...
+                    curator.repomgrcli,
+                    [
+                        '--debug',
+                        '--logfile', os.devnull,
+                        '--host', host,
+                        '--port', str(port),
+                        'create',
+                        'fs',
+                        '--repository', self.args['repository'],
+                        '--location', '/tmp'
```
untergeek: This should probably not be hard-coded to `/tmp` because it is not OS independent.  Perhaps a tmpdir provided by Python? 

```
import tempfile
tmpdir = tempfile.gettempdir()
```

test/integration/test_cli_commands.py
untergeek: `host` and `port` are defined by environment variables, so that's how that's taken care of.  Again, it's manually specified, and that's what we're looking for here.  It overrides the default, even if it overrides it with the same value.

test/integration/test_cli_commands.py
feltnerm: Good point,

One issue is that we need to know the tempdir of the _target_ OS, not our own (target: where ES is running).

untergeek: That's an excellent point.  But I'm willing to let this one fail in instances where there's no matchup, just in case.  Most of the time the instance used in testing is on the same host.

untergeek: Do document that inline, though.

feltnerm: I'll add a comment, definitely. Since we already have `self.args['location'] = tempfile.mkdtemp(suffix=dirname)` in `CuratorTestCase` it seems to me that I can just use `self.args['location']` in the actual test case.

untergeek:
Excellent.  You have two more edits to make.

1. Add yourself to the CONTRIBUTORS file.
2. Add a blurb to the Changelog.rst file.

feltnerm:
Done! 
Thanks for your assistance! You made it trivial to contribute my patch. A+++ would pull request again!

untergeek:
lgtm!

PR #383, opened 2015-05-21
There is no _status endpoint before 1.1.0

### LINKED ISSUE #379: Tests are failing because of API issues with older versions of ES
Specifically the 1.0_nightly test and the 1.0.3 test.

`TransportError(404, 'RepositoryMissingException[[_status] missing]')` is happening when `in_progress = client.snapshot.status()['snapshots']` is called, because no repository name is provided.

I will first test this by adding `repository='_all'`.  If that doesn't work, I may revert to using the named repository, though I want it to check against all.

untergeek:
This should put the final nail in the coffin.

fixes #379


untergeek:
YES! Finally fixed!

PR #414, opened 2015-06-24
Corrected two errors in the more complex example.

jrmycanady:
The filter for loop operated on the wrong variable name and the apply_filter method was not being called from the curator object.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -60,3 +60,4 @@ Contributors:
 * (pkr1234)
 * Mark Feltner (feltnerm)
 * William Jimenez (wjimenez5271)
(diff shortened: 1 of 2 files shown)


untergeek:
Thanks for catching that!

Feel free to add yourself to the CONTRIBUTORS file.  I'll be happy to merge when that's done!

untergeek:
And after you've signed the [contributor's agreement](http://www.elasticsearch.org/contributor-agreement/), too.

jrmycanady:
I've added myself to the CONTRIBUTORS file and signed the contributor's agreement. Let me know if there is anything else I need to do!

untergeek:
Thanks for making Curator more awesome!

Issue #420, opened 2015-06-26
curator v3.2.0 reporting successful indices deletion, but does not actually deleted them


TinLe:
I have a cluster with bunch of old indices that I want to prune, keeping only latest 7 days.  I run the command.

$ curator --port 9217 --timeout 900 delete indices --timestring "%Y.%m.%d" --older-than 7 --time-unit days
2015-06-25 16:54:49,038 INFO      Job starting: delete indices
2015-06-25 16:54:49,628 INFO      Pruning Kibana-related indices to prevent accidental deletion.
2015-06-25 16:54:49,628 WARNING   Very large list of indices.  Breaking it up into smaller chunks.
2015-06-25 16:54:49,629 INFO      Deleting indices as a batch operation:
.....bunch of outputs
2015-06-25 16:55:33,241 INFO      Job completed successfully.

But that is not true.   I checked and there are still many indices that were supposed to be deleted and were not.   So if I run the same command multiple times, eventually all the indices get deleted.

The output of multiple runs is fairly long, I'll put them in a gist here:
https://gist.github.com/TinLe/c64dbe495e0c2d228a6b

The last few runs correctly gave Error or Warnings, so that part is correct.   The error/warning were because the cluster just started creating new daily indices, so were busy with initializing shards (yellow status).   Once the cluster got done with the new indices, the deletion completed fine.   When I started the deletion, the cluster was relatively quiet and in green status.

The bug is that curator is reporting "Job completed successfully" when obviously that is not the case.

untergeek:
I appreciate that this is a bug for you.  This is _extremely_ unusual, except in cases where the cluster is heavily overburdened.  The `delete` API call [checks for exceptions](https://github.com/elastic/curator/blob/master/curator/api/delete.py#L16-L24), and if there are none, it returns True.  This is why it's considered a successful operation.  Elasticsearch itself is capable of deleting hundreds of indices simultaneously when not under egregious loads.  The test suite itself does this with many indices simultaneously.  If your cluster is failing to do this, you should look at the Elasticsearch logs to see what's happening.  Even the time elapsed in the log you provided suggests it sent all of the deletes in a 2 second window.

What happens if you run this with the `--debug` flag?  If you're trying to delete older than 7 days, why are you waiting until there are 5 indices of every kind, rather than running daily to prune the number of indices more regularly?  How many nodes in your cluster?  How many shards per node?  The count of indices you have is concerning and could be a contributing factor.

untergeek:
[...]

PR #423, opened 2015-06-26
Fix an edge case where deletes don't work properly

untergeek:
For some reason, deletes don't happen immediately in #420

This retries up to 3 times before giving a failing message.


untergeek:
I hate merge conflicts...

PR #443, opened 2015-07-19
Added allocation types: include and exclude

steffo:
Extended the ability to use both include and exclude for allocation strategies.

### DIFF
--- a/curator/api/allocation.py
+++ b/curator/api/allocation.py
@@ -3,12 +3,13 @@ import elasticsearch
 import logging
 logger = logging.getLogger(__name__)
 
-def apply_allocation_rule(client, indices, rule=None):
+def apply_allocation_rule(client, indices, rule=None, allocation_type='require' ):
     """
     Apply a required allocation rule to a list of indices.
 
     :arg client: The Elasticsearch client connection
     :arg indices: A list of indices to act on
+    :arg allocation_type: Which type of allocation to apply
     :arg rule: The routing allocation rule to apply, e.g. ``tag=ssd``.  Must be
         in the format of ``key=value``, and should match values declared on the
         correlating nodes in your cluster.
@@ -23,21 +24,25 @@ def apply_allocation_rule(client, indices, rule=None):
         return False
     key = rule.split('=')[0]
     value = rule.split('=')[1]
-    indices = prune_allocated(client, indices, key, value)
+    indices = prune_allocated(client, indices, key, value, allocation_type)
+
+    if allocation_type != 'require' and allocation_type != 'include' and allocation_type !='exclude':    
+        return False
+
     if not indices:
         logger.warn("No indices to act on.")
         return False
-    logger.info('Updating index setting index.routing.allocation.require.{0}={1}'.format(key,value))
+    logger.info('Updating index setting index.routing.allocation.{0}.{1}={2}'.format(allocation_type,key,value))
     try:
         client.indices.put_settings(index=to_csv(indices),
-            body='index.routing.allocation.require.{0}={1}'.format(key,value),
+            body='index.routing.allocation.'+allocation_type+'.{0}={1}'.format(key,value),
             )
         return True
     except:
         logger.error("Error in updating index settings with allocation rule.  Run with --debug flag and/or check Elasticsearch logs for more information.")
         return False
 
-def allocation(client, indices, rule=None):
+def allocation(client, indices, rule=None, allocation_type='require' ):
     """
     Helper method called by the CLI.
 
@@ -48,4 +53,4 @@ def allocation(client, indices, rule=None):
         correlating nodes in your cluster.
     :rtype: bool
     """
-    return apply_allocation_rule(client, indices, rule=rule)
+    return apply_allocation_rule(client, indices, rule=rule, allocation_type=allocation_type )
--- a/curator/api/allocation.py
+++ b/curator/api/allocation.py
@@ -9,10 +9,10 @@ def apply_allocation_rule(client, indices, rule=None, allocation_type='require' 
 
     :arg client: The Elasticsearch client connection
     :arg indices: A list of indices to act on
-    :arg allocation_type: Which type of allocation to apply
     :arg rule: The routing allocation rule to apply, e.g. ``tag=ssd``.  Must be
         in the format of ``key=value``, and should match values declared on the
         correlating nodes in your cluster.
+    :arg allocation_type: Type of allocation to apply
     :rtype: bool
 
     .. note::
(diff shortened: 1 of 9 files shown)


curator/api/allocation.py
```diff
@@ -23,21 +24,25 @@ def apply_allocation_rule(client, indices, rule=None):
...
+
     if not indices:
         logger.warn("No indices to act on.")
         return False
-    logger.info('Updating index setting index.routing.allocation.require.{0}={1}'.format(key,value))
+    logger.info('Updating index setting index.routing.allocation.{0}.{1}={2}'.format(allocation_type,key,value))
     try:
         client.indices.put_settings(index=to_csv(indices),
-            body='index.routing.allocation.require.{0}={1}'.format(key,value),
+            body='index.routing.allocation.'+allocation_type+'.{0}={1}'.format(key,value),
```
untergeek: Why not do the same thing here as in line 35?

```
body='index.routing.allocation.{0}.{1}={2}'.format(allocation_type,key,value),
```

curator/api/allocation.py
```diff
@@ -23,21 +24,25 @@ def apply_allocation_rule(client, indices, rule=None):
         return False
     key = rule.split('=')[0]
     value = rule.split('=')[1]
-    indices = prune_allocated(client, indices, key, value)
+    indices = prune_allocated(client, indices, key, value, allocation_type)
+
+    if allocation_type != 'require' and allocation_type != 'include' and allocation_type !='exclude':
```
untergeek: While this works, it's rather long.  Perhaps something more concise like this:

```
if allocation_type not in ['require', 'include', 'exclude']:
```

curator/api/allocation.py
```diff
@@ -23,21 +24,25 @@ def apply_allocation_rule(client, indices, rule=None):
         return False
     key = rule.split('=')[0]
     value = rule.split('=')[1]
-    indices = prune_allocated(client, indices, key, value)
+    indices = prune_allocated(client, indices, key, value, allocation_type)
+
+    if allocation_type != 'require' and allocation_type != 'include' and allocation_type !='exclude':    
+        return False
```
untergeek: I think an error log message here would be appropriate:

```
logger.error("{0} is an invalid allocation_type.  Must be one of 'require', 'include', 'exclude'.".format(allocation_type))
```

curator/api/allocation.py
```diff
@@ -48,4 +53,4 @@ def allocation(client, indices, rule=None):
         correlating nodes in your cluster.
```
untergeek: Please add `allocation_type` to the docs here.

curator/api/allocation.py
```diff
@@ -3,12 +3,13 @@
...
 logger = logging.getLogger(__name__)
 
-def apply_allocation_rule(client, indices, rule=None):
+def apply_allocation_rule(client, indices, rule=None, allocation_type='require' ):
     """
     Apply a required allocation rule to a list of indices.
 
     :arg client: The Elasticsearch client connection
     :arg indices: A list of indices to act on
+    :arg allocation_type: Which type of allocation to apply
```
untergeek: Please move this below `rule`, so they appear in the same order the arguments appear in the definition.

curator/api/utils.py
```diff
@@ -310,6 +310,7 @@ def prune_allocated(client, indices, key, value):
     :arg indices: A list of indices to act on
     :arg key: The allocation attribute to check for
     :arg value: The value to check for
+    :arg allocation_type: type of allocation
```
untergeek: Please try to follow the descriptive patterns and capitalization.  This could be something like "The allocation type to check for.  Must be one of 'require', 'include', or 'exclude'."

curator/api/utils.py
```diff
@@ -319,11 +320,11 @@ def prune_allocated(client, indices, key, value):
             index=idx,
         )
         try:
-            has_routing = settings[idx]['settings']['index']['routing']['allocation']['require'][key] == value
+            has_routing = settings[idx]['settings']['index']['routing']['allocation'][allocation_type][key] == value
         except KeyError:
             has_routing = False
         if has_routing:
-            logger.debug('Skipping index {0}: Already has allocation rule {1} applied.'.format(idx, key + "=" + value))
+            logger.debug('Skipping index {0}: Already has allocation rule {1} applied for type {2}.'.format(idx, key + "=" + value, allocation_type))
```
untergeek: This doesn't quite fit the syntax of `type.rule=value`.  Perhaps "Allocation rule {1} is already applied for type {2}."  

Thoughts?

curator/cli/allocation.py
```diff
@@ -8,11 +8,19 @@
...
 @click.pass_context
-def allocation(ctx, rule):
+def allocation(ctx, rule, type):
     """Index Allocation"""
     if not rule:
         click.echo('{0}'.format(ctx.get_help()))
         click.echo(click.style('Missing required parameter --rule', fg='red', bold=True))
         sys.exit(1)
+
+    if type != 'require' and type != 'include' and type !='exclude':
```
untergeek: As with the other, this could be more cleanly written as:

```
if type not in ['require', 'include', 'exclude']:
```

curator/cli/allocation.py
```diff
@@ -8,11 +8,19 @@
...
+def allocation(ctx, rule, type):
     """Index Allocation"""
     if not rule:
         click.echo('{0}'.format(ctx.get_help()))
         click.echo(click.style('Missing required parameter --rule', fg='red', bold=True))
         sys.exit(1)
+
+    if type != 'require' and type != 'include' and type !='exclude':    
+		#click.echo('{1}'.format(ctx.get_help()))
+        click.echo(click.style('--type can only one of: require, include  exclude', fg='red', bold=True))
```
untergeek: Add `be`, e.g. '--type can only _be_ one of'...

test/unit/test_api_commands.py
```diff
@@ -175,28 +175,42 @@ def test_allocation_rule_positive(self):
...
         client = Mock()
-        self.assertFalse(curator.apply_allocation_rule(client, [], rule="foo=bar"))
+        self.assertFalse(curator.apply_allocation_rule(client, [], rule="foo=bar" ))
     def test_allocation_positive(self):
         client = Mock()
         client.cluster.state.return_value = open_index
         client.indices.get_settings.return_value = allocation_out
         client.indices.put_settings.return_value = None
-        self.assertTrue(curator.allocation(client, named_index, rule="foo=bar"))
+        self.assertTrue(curator.allocation(client, named_index, rule="foo=bar" ))
```
untergeek: Why do all of these lines have an extra space at the end after `"foo=bar"`?

untergeek:
Thanks for adding this functionality.  Honestly, I'm kind of surprised that nobody has requested it before.

Can you respond to the suggestions I've made, and perhaps look into adding some integration tests as well?  If not, I will add those later, but I definitely want an integration test per type.

steffo:
Hi, 
Added changes a per review. Integration test are underway. I don't understand why check is failing now. Working on it.

untergeek:
Please be sure to also:

* Add yourself to the CONTRIBUTORS file
* Add something to the Changelog.rst file

untergeek:
Sorry for the delay.  Was in Portland for OSCON and training.  I've been catching up ever since.

PR #470, opened 2015-08-28
Testing some testing regressions

untergeek:
Something is different with elasticsearch-py v1.6.0.

Don't know what yet, as it passes on my machine.  This is a first stab
at deciphering it.


untergeek:
Nevermind.  It appears these regressions are actually problems with Jenkins itself, not my code.

PR #487, opened 2015-10-06
#476: select only closed indices while filtering

### LINKED ISSUE #476: [Feature] Filtering by "closed"
I'd like to delete old closed indices. I can filter by time, by not by "closed" state. Couldn't find, how to do so. Is it possible? Would it make sense to have such a filter?

Basster:
These are my first steps in python,so please be patient but I needed this filter urgently.

This should hopefully fix #476.

### DIFF
--- a/curator/api/utils.py
+++ b/curator/api/utils.py
@@ -20,13 +20,10 @@ def get_alias(client, alias):
         logger.error('Unable to find alias {0}.'.format(alias))
         return False
 
-def get_indices(client, closed_only=False):
+def get_indices(client):
     try:
-        wildcards = 'closed'
-        if closed_only is False:
-            wildcards = 'open,' + wildcards
         indices = list(client.indices.get_settings(
-            index='_all', params={'expand_wildcards': wildcards}))
+            index='_all', params={'expand_wildcards': 'open,closed'}))
         logger.debug("All indices: {0}".format(indices))
         return indices
     except Exception:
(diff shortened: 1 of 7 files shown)


untergeek:
Thanks for diving in!  Your python looks just fine, and the test addition looks great.  :smile: 

Looking at what you're trying to accomplish, though, I'd prefer to see this applied as a filter.  I really don't want to have the `get_indices` method changed, even though what you've created does work.  It's an opportunity to extend the API that is behind my motivation.

If you look at https://github.com/elastic/curator/blob/master/curator/api/utils.py#L287-L302, you'll see that there's already a `prune_closed` method.  It should be pretty simple to clone that, but make it `prune_opened`.  **This extends the API,** which is why I'd prefer it if you went in that direction.

Remove the `not` from https://github.com/elastic/curator/blob/master/curator/api/utils.py#L298, and change `Closed` to `Opened` in https://github.com/elastic/curator/blob/master/curator/api/utils.py#L301.

You won't have to change the config option you created.  You'd only need to put an if statement after https://github.com/elastic/curator/blob/master/curator/cli/index_selection.py#L84-L87 to filter out the "open" indices.  Something like this, perhaps:

```
if closed_only:
    logger.info("Pruning open indices, leaving only closed indices.")
    working_list = prune_opened(client, working_list)
```

How does that sound?

Basster:
Sounds good. I've updated the code accordingly.

curator/api/utils.py
```diff
@@ -292,13 +292,27 @@ def prune_closed(client, indices):
...
+    """
+    Return list of indices that are not open.
+
+    :arg client: The Elasticsearch client connection
+    :arg indices: A list of indices to act on
+    :rtype: list
+    """
+    return prune_open_or_closed(client, indices, append_closed=True)
+
+def prune_open_or_closed(client, indices, append_closed):
```
untergeek: This was a clever approach!  

I'd prefer to see `append_closed` as a keyword argument with a default, e.g. `append_closed=False`, rather than as a positional argument.  Also, I feel that `return_closed` is better semantically than `append_closed`, even if appending is what is happening.  We want the method to _return_ the closed values.

untergeek: You know what?  This is fine.  Disregard my previous comment.  We'll stick with this.

untergeek:
@Basster 

The only things I need from you now are to update the CONTRIBUTORS file and Changelog.rst.

Update the Changelog with what you've changed, and add yourself to CONTRIBUTORS.

Basster:
*done* and thank you for your feedback.

Issue #492, opened 2015-10-16
Documentation should point to elastic.co


jgoodall:
On [installation page](https://www.elastic.co/guide/en/elasticsearch/client/curator/current/installation.html), the repository install instructions (APT, YUM) refer to `elasticsearch.org`, but should probably be updated to `elastic.co` to match the instructions for elasticsearch itself.

untergeek:
packages.elasticsearch.org and packages.elastic.co have different IPs:

```
» host packages.elasticsearch.org
packages.elasticsearch.org is an alias for dualstack.downloadlb-1287877610.us-east-1.elb.amazonaws.com.
dualstack.downloadlb-1287877610.us-east-1.elb.amazonaws.com has address 184.72.222.192
dualstack.downloadlb-1287877610.us-east-1.elb.amazonaws.com has address 107.22.185.174
dualstack.downloadlb-1287877610.us-east-1.elb.amazonaws.com has address 184.72.233.150
dualstack.downloadlb-1287877610.us-east-1.elb.amazonaws.com has IPv6 address 2406:da00:ff00::b848:e996
dualstack.downloadlb-1287877610.us-east-1.elb.amazonaws.com has IPv6 address 2406:da00:ff00::6b16:b9ae
dualstack.downloadlb-1287877610.us-east-1.elb.amazonaws.com has IPv6 address 2406:da00:ff00::b848:dec0

» host packages.elastic.co
packages.elastic.co is an alias for dualstack.download-colb-770446651.us-east-1.elb.amazonaws.com.
dualstack.download-colb-770446651.us-east-1.elb.amazonaws.com has address 184.73.247.78
dualstack.download-colb-770446651.us-east-1.elb.amazonaws.com has address 184.73.196.98
dualstack.download-colb-770446651.us-east-1.elb.amazonaws.com has address 50.19.101.59
dualstack.download-colb-770446651.us-east-1.elb.amazonaws.com has IPv6 address 2406:da00:ff00::b849:f74e
dualstack.download-colb-770446651.us-east-1.elb.amazonaws.com has IPv6 address 2406:da00:ff00::3213:653b
dualstack.download-colb-770446651.us-east-1.elb.amazonaws.com has IPv6 address 2406:da00:ff00::b849:c462

```

We're transitioning to a new repository.  Until that's complete, this will need to remain as-is.

jgoodall:
Will both continue to work?

untergeek:
Yes.

untergeek:
[...]

PR #493, opened 2015-10-16
Use packages.elastic.co instead

untergeek:
Since they use the same AWS bucket, it's a legitimate change.  It's also non-breaking as the other will persist for a while.


untergeek:
fixes #492

PR #494, opened 2015-10-19
Fix/438

### LINKED ISSUE #438: When using logstash log format, all output should respect that
Currently, there appears to be output that doesn't respect `--logformat logstash` — for example, if I run:

`curator --logformat logstash delete indices --older-than 10 --time-unit days --timestring %Y.%m.%d`

and there aren't any indices older than 10 days, I get the "Job starting" and then the "No indices matched provided args" lines output in properly-formatted logstash JSON, but then I get the "No indices matched provided args" output AGAIN as plain text.

This means that if I pipe the curator output to logstash, the JSON-formatted lines get consumed properly, but the plain-text one causes logstash to hiccup and report a JSON parse failure.

untergeek:
Add --quiet output

(also bump dependency to elasticsearch-py 1.8.0)

fixes #438

### DIFF
--- a/curator/cli/alias.py
+++ b/curator/cli/alias.py
@@ -12,7 +12,8 @@ logger = logging.getLogger(__name__)
 def alias(ctx, name, remove):
     """Index Aliasing"""
     if not name:
-        click.echo('{0}'.format(ctx.get_help()))
-        click.echo(click.style('Missing required parameter --name', fg='red', bold=True))
+        msgout('{0}'.format(ctx.get_help()), quiet=ctx.parent.params['quiet'])
+        logger.error('Missing required parameter --name')
+        msgout('Missing required parameter --name', error=True, quiet=ctx.parent.params['quiet'])
         sys.exit(1)
 alias.add_command(indices)
--- a/curator/cli/allocation.py
+++ b/curator/cli/allocation.py
@@ -14,12 +14,14 @@ logger = logging.getLogger(__name__)
 def allocation(ctx, rule, type):
     """Index Allocation"""
     if not rule:
-        click.echo('{0}'.format(ctx.get_help()))
-        click.echo(click.style('Missing required parameter --rule', fg='red', bold=True))
+        msgout('{0}'.format(ctx.get_help()), quiet=ctx.parent.params['quiet'])
+        logger.error('Missing required parameter --rule')
+        msgout('Missing required parameter --rule', error=True, quiet=ctx.parent.params['quiet'])
         sys.exit(1)
 
     if type not in ['require', 'include', 'exclude']:
-        click.echo(click.style('--type can only be one of: require, include  exclude', fg='red', bold=True))
+        logger.error('--type can only be one of: require, include  exclude')
+        msgout('--type can only be one of: require, include  exclude', error=True, quiet=ctx.parent.params['quiet'])
         sys.exit(1)
 
 allocation.add_command(indices)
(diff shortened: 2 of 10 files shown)


cameronkerrnz:
After upgrading curator to version 3.4.0 yesterday and adding --quiet, I still get email from cron (although I don't get output when running the same script from the command-line).

~~~
/etc/cron.hourly/elk-purge-old-logs:

No indices matched provided args: {'regex': '^nxlog', 'index': (), 'suffix': None, 'newer_than': None, 'closed_only': False, 'prefix': None, 'time_unit': 'days', 'timestring': '%Y.%m.%d', 'exclude': (), 'older_than': 8, 'all_indices': False}
~~~

~~~
# cat /etc/cron.hourly/elk-purge-old-logs
#!/bin/bash

curator="curator --quiet --loglevel ERROR"
days="--time-unit days --timestring %Y.%m.%d"

basename_X_replicated_Ndays_deleted_after_Mdays()
{
  $curator replicas --count 0 indices --regex "^${1}" --older-than "${2}" $days
  $curator delete indices --regex "^${1}" --older-than "${3}" $days
}

$curator show indices --timestring %Y.%m.%d | sed -e 's/-....\...\...//' | sort -u | while read index_basename
do
  case "${index_basename}" in
    active_directory)
      basename_X_replicated_Ndays_deleted_after_Mdays "${index_basename}" 4 8
      ;;
    exchange)
      basename_X_replicated_Ndays_deleted_after_Mdays "${index_basename}" 4 8
      ;;
    *)
      basename_X_replicated_Ndays_deleted_after_Mdays "${index_basename}" 7 31
      ;;
  esac
done

exit 0
~~~

untergeek:
I should make it more clear that `--quiet` isn't the same thing as no output.  The `--quiet` flag suppresses stdout output, but not log output.  If you do not specify a log file, then Curator defaults to use stdout as its logging target.  This is likely what you're experiencing: stdout log output.

If you want it to be completely quiet, you will need to use a `--logfile`.  Without it, you'll still be logging to stdout, hence output.

Why would I have both stdout _and_ logfile output?  So that users will get some messages at the command-line no matter what (except in the case of the new `--quiet` flag).

untergeek:
@cameronkerrnz please see the above comment

cameronkerrnz:
Thanks for the clarification, although I have some feedback about that.

If there is a reasonable use-case for making "No indices matched provided args..." an ERROR level message, then I think that behaviour should be configurable. As an idempotent operation (ie. we're promising to delete indices older than some criteria), the fact that there were no such indices found in this particular iteration isn't an error. Perhaps the log-level of that message ought to be an INFO-level message. Otherwise, I either generate useless logs, or have to filter out this particular message so it doesn't generate useless cron mail (which is worse because then it tends to get filtered and ignored).

I still don't understand though, why the log output is generated when run under cron, but not when run manually. Running it with stdin being /dev/null doesn't reproduce that same behaviour.

I have now updated my script to be:

~~~
#!/bin/bash

days="--time-unit days --timestring %Y.%m.%d"
curator="/usr/local/bin/curator --quiet --loglevel ERROR"

curator_wrapper() {
  $curator "$@" 2>&1 | grep -v -e '^No indices matched provided args'
}


basename_X_replicated_Ndays_deleted_after_Mdays()
{
  curator_wrapper replicas --count 0 indices --regex "^${1}" --older-than "${2}" $days
  curator_wrapper delete indices --regex "^${1}" --older-than "${3}" $days
}

$curator show indices --timestring %Y.%m.%d | sed -e 's/-....\...\...//' | sort -u | while read index_basename
do
  case "${index_basename}" in
    active_directory)
      basename_X_replicated_Ndays_deleted_after_Mdays "${index_basename}" 4 8
      ;;
    exchange)
      basename_X_replicated_Ndays_deleted_after_Mdays "${index_basename}" 4 8
      ;;
    *)
      basename_X_replicated_Ndays_deleted_after_Mdays "${index_basename}" 7 31
      ;;
  esac
done

exit 0
~~~

This seems to work as expected.

Thanks for your work on curator.

Cheers,
Cameron

Issue #491, opened 2015-10-13
Compatibility with AWS ES Service


univerio:
Curator at the moment is sightly incompatible with the AWS hosted ElasticSearch Service when using the `optimize` command because AWS ES does not expose the `/_cluster/state/metadata` endpoint (returns 401), which curator uses to determine if an index is closed.

I understand that perhaps the solution is to ask AWS to expose the endpoint. I've made a thread about it [here](https://forums.aws.amazon.com/thread.jspa?threadID=217758).

On the curator side, though, perhaps it could warn when it can't determine if a index is closed instead of crashing with an exception. Presumably it would just fail to optimize anyway if it's closed?

untergeek:
@univerio I'm sorry to hear you're having difficulty with AWS ES.  An endpoint-limited version of ES is not a scenario I could have planned for.  Curator was designed to run with full-service versions of Elasticsearch because there was never an expectation there would ever be anything but.

Curator depends on the `index_closed` method in more places than just the `optimize` method, and is also referenced by the `prune_closed` method.  I can't just turn off or ignore exceptions for this check without repercussions.  These methods were added _because_ of the program-crashing results of trying to perform operations on closed indices. The only 2 things you can do to closed indices are open and delete (and delete actually briefly opens the index—internally to Elasticsearch—before deleting).  While I appreciate that AWS ES does not allow users to even close indices, I'm concerned that the path you recommend—making Curator regard the exception as only a warning—could have some unintended consequences for Elasticsearch deployments which can close indices.

If you (or someone else) were to code up a way around this, and add tests to guarantee reverse compatibility with all versions of Elasticsearch from 1.0.x - 2.0.x, I'd be willing to merge that in, if it passes code review.

univerio:
That's unfortunate. What do you think a workaround would look like? A command-line flag that turns on AWS ES-specific behavior? Or trying to automatically detect that the ES instance is from AWS?

untergeek:
I suppose you could parse that out.  I'd think it'd be easier to use the `_cat` API with the JSON output.  The AWS ES version is 1.5+, and the _cat API will show index state (open or closed), but only in Elasticsearch versions > 1.4.

Modifying the `index_closed` function to do an ES version check (Curator does this in several places already), then use the `_cat` API first if an acceptable version of ES is found would solve this in a less fragile way.

univerio:
Great, I'll prepare a pull request.

[...]

PR #499, opened 2015-10-22
Make index_closed use /_cat/indices instead of /_cluster/state for ES > 1.5.0

univerio:
#491

### DIFF
--- a/curator/api/utils.py
+++ b/curator/api/utils.py
@@ -4,6 +4,7 @@ import time
 import re
 import sys
 import logging
+import json
 logger = logging.getLogger(__name__)
 
 def get_alias(client, alias):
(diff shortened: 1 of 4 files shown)


untergeek:
Vanilla 1.5.1 returns a list.  The elasticsearch module does the JSON handling.  AWS Elasticsearch is seriously screwed up if you can't get a proper response (text/json)  for the _cat API such that the Elasticsearch module works.

curator/api/utils.py
```diff
@@ -88,11 +89,18 @@ def index_closed(client, index_name):
...
     :rtype: bool
     """
-    index_metadata = client.cluster.state(
-        index=index_name,
-        metric='metadata',
-    )
-    return index_metadata['metadata']['indices'][index_name]['state'] == 'close'
+    if get_version(client) >= (1, 5, 0):
+        indices = client.cat.indices(index=index_name, format='json', h='status')
+        if not isinstance(indices, list):  # content-type: text/plain
```
untergeek: Please add a comment here to explain that this workaround is for AWS Elasticsearch, that it is using `text/plain` to respond.

untergeek: I'll do this myself.  Tests are passing for all versions of Elasticsearch tested, from 1.0.3 -> 2.0.0.rc1

PR #507, opened 2015-11-10
return False when failing to get _all snapshots

garyelephant:
`if name in all_snaps` will raise `TypeError` when `get_snapshots` returns  False.

```
TypeError: argument of type 'bool' is not iterable
```


untergeek:
Can you add a test case for this?

garyelephant:
Sorry, I have no idea to write test code, although I read the code in test folder,
Could you please tell me how to write test code in such condition:
https://github.com/garyelephant/curator/blob/master/curator/api/utils.py#L253

untergeek:
No worries.  I don't think this needs a test case.  Thanks for adding to Curator.

PR #509, opened 2015-11-16
fix doc inconsistency

pmoust:


untergeek:
Thanks for catching this :smile:

PR #513, opened 2015-11-20
Update show.asciidoc

pbamba:
The show-indices command doesn't seem to exists.
The fixed command works.
It seems to be a typo.


untergeek:
Thanks for catching the typo.

PR #517, opened 2015-12-02
support to sort index names starting on given character position when deleting by disk space

imreFitos:
curator delete --disk-space sorts alphabetically by default. This creates problems if we have multiple index names with timestamps, e.g.:

apache-2015.12.02 
apache-2015.12.01
nginx-2015.12.02
nginx-2015.12.01

curator would potentially delete both the nginx logs instead of the oldest apache and nginx log.

This patch adds a "position" parameter to the delete command that allows one to specify at what character the sort should start. It also allows for negative positions that start from the end of the index name:

    curator delete --disk-space 1 --position -10 

Sort from the tenth character counted from the end of the index name, e.g. "apache-2015.12.01" would be sorted as "2015.12.01"

### DIFF
--- a/curator/api/filter.py
+++ b/curator/api/filter.py
@@ -301,7 +301,7 @@ def timestamp_check(timestamp, timestring=None, time_unit=None,
                         timestamp, method.replace('_', ' '), value, time_unit))
     return False
 
-def filter_by_space(client, indices, disk_space=None, reverse=True):
+def filter_by_space(client, indices, disk_space=None, reverse=True, position=0):
     """
     Remove indices from the provided list of indices based on space consumed,
     sorted reverse-alphabetically by default.  If you set `reverse` to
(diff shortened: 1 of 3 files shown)


untergeek:
I apologize, but in its current state, I will not be merging this pull request.

I appreciate the effort that went into this.  Unfortunately, it's very limited in its current utility, in that it only filters by position from the right.  This works for your use case, but it does not solve the bigger problem of sorting by date after sorting by space.  I'd much rather see a change to `delete --disk-space` that adds flags and first sorts by date, and then sorts by disk space.  It's a much bigger problem than it first appears (which is why it has not been given much love).

Fortunately, there are regexes and methods already in the API that could hypothetically be applied to the indices to sort them by age before their sizes are calculated.  It would require a partial reimplementation of the filtering criteria in https://github.com/elastic/curator/blob/master/curator/cli/index_selection.py, but it could be done.  It could even take options given in the `indices` sub-command to help with the filtering.

imreFitos:
hey thank you for the prompt comment!

Could you describe your scenario in more detail? Would your date be retrieved from the index names, or from somewhere else? What are your indexes named? How would you like the arguments to look like?

untergeek:
Currently, you can do `delete --disk-space` and then `indices` to further filter.  I propose adding an extra flag to go along with `--disk-space` that would sort by date.  Date bits would have to go into the `indices` section, but the filter-by-space method could theoretically extract that data from the click context, and use  it as well.

That's just the rough view at this point.

imreFitos:
So where would you get the dates from? the index names or from some other source? And what would you like the new flag to be called?

untergeek:
I'm not sure it will be easy.  Perhaps it will require re-using the --timestring flag, though I would hope it could be passed along from the context.  It may require extra args and pass-throughs to accomplish this, however.

untergeek:
This is obviated by the pending release of Curator 4.0.

See the changes in #595 and #596

PR #522, opened 2015-12-07
Added newest and oldest command line options

vvanholl:
--newest : integer>=0, returns the n newest items
needs the --timestring option

--oldest : integer>=0 returns the n oldest items
needs the --timestring option

### DIFF
--- a/curator/api/filter.py
+++ b/curator/api/filter.py
@@ -7,6 +7,8 @@ logger = logging.getLogger(__name__)
 
 REGEX_MAP = {
     'timestring': r'^.*{0}.*$',
+    'newest': r'(?P<date>{0})',
+    'oldest': r'(?P<date>{0})',
     'newer_than': r'(?P<date>{0})',
     'older_than': r'(?P<date>{0})',
     'prefix': r'^{0}.*$',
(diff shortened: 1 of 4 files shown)


vvanholl:
It seems there is a package dependency error during the CI check:

pkg_resources.VersionConflict: (mock 1.3.0 (/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages), Requirement.parse('mock==1.0.1'))

Could you help me please ?

untergeek:
I don't rely on Travis, but on our internal CI

vvanholl:
OK, this means that my pull request will be still examined ?

untergeek:
I am curious about the use case.  Why does a fixed number of newest/oldest matter?

untergeek:
What I mean to say, is when will you delete the _n_ oldest, when theoretically, that number could include today's index?  How is this an improvement over `--newer-than` and `--older-than`?

Please sell me on why this functionality ought to be included.

vvanholl:
Hi,

I think this feature could be useful for instance when you need to make some cleanup on your disk by deleting the oldest indices first. 

Vincent

untergeek:
I'm still not sold.  While I can appreciate your use case, this adds complexity and more flags to an already large option tree but adds very little additional functionality.

corey-hammerton:
Especially when you have a scenario like retiring logstash data in a 30 day cycle then gets changed to a 21 day cycle. Using the ``--oldest`` flag won't help in that case.

untergeek:
Thank you for contributing to Curator.  While I appreciate the thought and effort that went into submitting this pull request, I have decided not to merge this code as it is too similar to existing functionality, and adds unnecessary complexity.

Please continue to support this project, and other open source projects, with your time, ideas, code, and comments.  Thank you!

Issue #540, opened 2016-01-11
Curator returns non-zero on non-master nodes


wkruse:
I have a systemd timer unit, that runs curator daily on all nodes in the Elasticsearch cluster (automated deployment). The indices should be deleted on master only.

```
curator --loglevel WARN --master-only --host xxx delete indices --prefix logstash --time-unit weeks --older-than 2 --timestring '%Y.%m.%d'
```

Curator returns non-zero on non-master nodes which leads to failed unit. I've expected it to return zero. Is it a bug?

untergeek:
It _was not_ a bug, but was to report a known error code for this case:

https://github.com/elastic/curator/blob/master/curator/cli/utils.py#L96-L102

```python
def check_master(client, master_only=False):
    """
    Check if master node.  If not, exit with error code
    """
    if master_only and not is_master_node(client):
        logger.info('Master-only flag detected. Connected to non-master node. Aborting.')
        sys.exit(9)
```

I see no problem changing this to exit with a 0 code, if you need this.  Feel free to submit a pull request with that change.  If you do, be sure to also update the Changelog with a blurb about what's been updated.

untergeek:
Closing via #541

sagh0900:
# **curator v3.5.1: (with SSL) [Elasticsearch 2.3.4(secured, not running behind any proxy), Search-Guard-2]**

I'm running curator from elected master-node:

When I run command with out master-only flag, it return output though with exceptions.
curator --loglevel DEBUG --use_ssl --host 192.168.0.1 --http_auth admin:admin --certificate "/usr/local/bin/root-ca.pem" --port 9200 show indices --all-indices

**output:** 

> INFO      Matching indices:

> /usr/lib/python2.7/site-packages/urllib3/connectionpool.py:770: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
>   InsecureRequestWarning)
> monitoring-2016.08.09
> /usr/lib/python2.7/site-packages/urllib3/connectionpool.py:770: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
>   InsecureRequestWarning)
> **searchguard**

**But when I include, master-only flag, it says connection failure, non-master node detected.**

[...]

PR #541, opened 2016-01-12
Change --master-only exit code to 0 when connected to non-master node

wkruse:
#540


untergeek:
I forgot to have you add yourself to the Contributors file.  Can you do that, too?

untergeek:
Thanks for this!

PR #568, opened 2016-02-16
Remove print() in setup.py as it breaks commands like "python setup.py --url"

thib-ack:
Hi,
We are using some standard commands to fetch data from the setup.py itself like 
$ python setup.py --url
$ python setup.py --description
This is broken since the setup.py is printing stuff when the cx_Freeze import fails.

I agree it's nice to know the import failed, but I think it should be printed somewhere else ?

Thanks,

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,6 @@ try:
     )
     ### end cx_Freeze ###
 except ImportError:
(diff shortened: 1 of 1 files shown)


untergeek:
Please sign the CLA. I will merge after that.

thib-ack:
Hi,
I just signed it.
Thanks,

PR #588, opened 2016-04-11
Define what the error level can be.

choffee:


choffee:
You can have that for free. I don't think it's worth signing up to the CLA for.

john

PR #612, opened 2016-04-26
Add more explicit doc for the direction filter element

Gueust:
The previous sentence "This setting is used to determine whether indices or snapshots are older or younger than the reference point in time determined by unit, unit_count, and optionally, epoch." was ambiguous, not saying if we were talking about the selected or the not selected indices.

Edit: I just did sign the CLA. Maybe it takes some time to change the status.

### DIFF
--- a/docs/asciidoc/filter_elements.asciidoc
+++ b/docs/asciidoc/filter_elements.asciidoc
@@ -9,10 +9,11 @@ NOTE: This setting is only used with the <<filtertype_age,age>> filtertype, or +
 with the <<filtertype_space,space>> filtertype if <<fe_use_age,use_age>> is
 set to `True`.
 
-This setting must be either `older` or `younger`.  This setting is used to
-determine whether indices or snapshots are `older` or `younger` than the
-reference point in time determined by <<fe_unit,unit>>,
-<<fe_unit_count,unit_count>>, and optionally, <<fe_epoch,epoch>>.
+This setting must be either `older` or `younger`. This setting is used to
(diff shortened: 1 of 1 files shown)


Gueust:
I just did sign the CLA. Maybe it takes some time to change the status.

Gueust:
The relation between the 'direction' and 'use_age' filter elements is unclear to me.

The test for the 'use_age' states:
````
n other words, it sorts all indices in the list by age, then starts adding the space consumed by each index, beginning from the youngest. Once the value of disk_space is reached, all remaining indices, which are the oldest, will be selected, omitting all of the younger indices.
````

What I understand it that the direction decides whether we are looking at older indices or younger indices compared to  the reference point. Among these indices, 'use_age' will always try to select the oldest (leaving the youngest unselected).

In that case, an even more accurate sentence for the PR could be:

This setting must be either `older` or `younger`. This setting is used to determine whether the filter should select **among the **indices or snapshots that are `older` or `younger` than the reference point in time determined by  <<fe_unit,unit>>, <<fe_unit_count,unit_count>>, and optionally,
<<fe_epoch,epoch>>.

untergeek:
There is no relationship between `direction` and `use_age`.  The space filter does not use the `direction` filter element, as space is _always_ calculated based on keeping the youngest indices, and deleting the oldest indices based on space consumption.  I'm working on better wording and explanation for these filters, so thank you for the pull request and suggestion.  It's a work in progress, while it's not yet fully released.

PR #627, opened 2016-05-05
fix error in example usage

fandrews:


untergeek:
Thanks for catching this.  It was so much documentation, I'm sure there are tiny mistakes like that in other places too 😦

PR #628, opened 2016-05-06
fix error message for missing timestring in filter_by_age

fandrews:


untergeek:
Thanks again.  By all means, add yourself to the Contributors file in another PR.

PR #662, opened 2016-06-27
Fix logformat json/logstash to send UTC offset

### LINKED ISSUE #661: Logging to JSON the field @timestamp overwrites the indexed time when using the logstash json filter
Logging to JSON the field @timestamp overwrites the indexed time when using the logstash json filter. 

```
{"@timestamp": "2016-06-25T14:03:58.179Z", "function": "__not_actionable", "linenum": 39, "loglevel": "INFO", "message": "Index .marvel-es-1-2016.06.22 is not actionable, removing from list.", "name": "curator.indexlist"}
```

I've tried to use remove_field to prevent it from happening but all i get in logstash is...

```
LogStash::Error: Unable to format in string "YYYY.MM.dd", @timestamp field not found
```

This is my filter...

```
filter {
    if [type] == "elasticsearch-curator" {
        json {
            source => "message"
            remove_field => [ "@timestamp" ]
        }
        mutate {
            rename => { "loglevel" => "level" }
        }
    }
}
```

Would it be possible to not use use the special @timestamp as a field name for your logs? 

I beleive the line of code that would need to be changed is [here](https://github.com/elastic/curator/blob/161aa69323213f77bdf008518773dc6d7c81d1a1/curator/logtools.py#L21)

untergeek:
The `@timestamp` field will now look like this:

```
{"@timestamp": "2016-06-27T11:07:37.906-0600", ... }
```

Hattip @ryancurrah

fixes #661

### DIFF
--- a/curator/logtools.py
+++ b/curator/logtools.py
@@ -1,6 +1,7 @@
 import sys
 import json
 import logging
+import time
 
 class LogstashFormatter(logging.Formatter):
     # The LogRecord attributes we want to carry over to the Logstash message,
@@ -15,6 +16,7 @@ class LogstashFormatter(logging.Formatter):
     #     return time.gmtime(timevalue)
 
     def format(self, record):
+        self.converter = time.gmtime
         timestamp = '%s.%03dZ' % (
             self.formatTime(record, datefmt='%Y-%m-%dT%H:%M:%S'), record.msecs)
         result = {'message': record.getMessage(),
--- a/docs/Changelog.rst
+++ b/docs/Changelog.rst
@@ -3,6 +3,14 @@
 Changelog
 =========
 
+4.0.1 (? ? ?)
+-------------
+
+**Bug Fixes**
+
+  * Coerce Logstash/JSON logformat type timestamp value to always use UTC.
+    #661 (untergeek)
+
 4.0.0 (24 June 2016)
 --------------------
 


ryancurrah:
I just patched this and while the formatting did change as expected, the timezone is reporting as UTC (+0000) so it is still broken. I did some googling to figure out how get the correct timezone seems its common issue. Stackoverflow article here shows you could just log everything in GMT/UTC timezone to simplify things or another answer use an external library like the common pytz module [Python logging module emits wrong timezone information](http://stackoverflow.com/questions/27858539/python-logging-module-emits-wrong-timezone-information). As it seems the logging function formatTime uses the time module which has poor support for timezone awareness.

```json
{"@timestamp": "2016-06-28T12:39:33.977+0000", "function": "cli", "linenum": 236, "loglevel": "ERROR", "message": "Unable to complete action \"delete_indices\".  No actionable items in list: <class 'curator.exceptions.NoIndices'>", "name": "curator.cli"}
```

Printed the date on server

```bash
[root@HOSTNAMEREDACTED ~]# date
Tue Jun 28 13:26:15 CDT 2016
```

ryancurrah:
Just to add another comment I did get it to work by forcing to log in UTC time..

So on top of your changes I added...

Imported time module...

```python
import time
```

Set logging converter attribute to a function with the same signature as time.gmtime right above your timestamp variable

```python
self.converter = time.gmtime
```

And ran Curator... the logs are now in proper UTC time...

```json
{"@timestamp": "2016-06-28T18:50:17.240+0000", "function": "cli", "linenum": 236, "loglevel": "ERROR", "message": "Unable to complete action \"delete_indices\".  No actionable items in list: <class 'curator.exceptions.NoIndices'>", "name": "curator.cli"}
```

Kibana/Elasticsearch is happy now too...

screen shot 2016-06-28 at 2 51 30 pm

untergeek:
Interesting that your installation didn't pick up the timezone automatically.  Good to know.  This will make the patch very small.

untergeek:
See if the changes I just pushed meet with your approval.  There's no need to send an offset if we're always sending in UTC (Zulu) time, so I reverted back to the older **Z** notation.

untergeek:
I'm going to merge this now, as I have some other bits I want to push in another PR, and I think I've addressed your concerns.

ryancurrah:
@untergeek sorry the delayed response, i applied the changes you made (reverting back the timestamp) and tested it is working fine. Thanks for the quick turn around!

PR #690, opened 2016-07-14
Added colon after "field"

ynux:
else yaml complains (if you blindly copy paste this)
curator.exceptions.ConfigurationError


untergeek:
Thanks!

PR #691, opened 2016-07-14
Add note to settings available in all actions

dizzler:
Added notes match the existing note for timeout_override, which is one of three settings available for all actions in 4.0.0 version of curator.


untergeek:
Thanks for updating the docs!

PR #710, opened 2016-07-28
add pyyaml dependency

dtrv:
as mentioned in requirements.txt within curator package


untergeek:
Thanks for adding this!

PR #750, opened 2016-09-01
Add support for environment variables

### LINKED ISSUE #697: Use environment variables with Curator v4
I'm using 3.5 and it accepts flags. this works well with environment variables such as

`--port ${PORT}`, etc.

It's unclear in the v4 docs on how to do this with the change to a configuration file.
The config file has everything I need, but it doesn't show if it accepts environment variables/overrides. Is it possible without modifying the file to do string replacements?

untergeek:
There were quite a few changes that had to go in, including further strictures on the schema, namely `Coerce` to `int` and `float`, since environment variables are all strings.

More tests were added, and the Changelog updated.

Added Elasticsearch 2.4.0 to the list of instances to be tested.

closes #697

### DIFF
--- a/curator/validators/filters.py
+++ b/curator/validators/filters.py
@@ -21,22 +21,22 @@ def structure():
     retval = {
         Optional('aliases'): Any(str, [str]),
         Optional('allocation_type'): str,
-        Optional('count'): int,
+        Optional('count'): Coerce(int),
         Optional('direction'): str,
         Optional('disk_space'): float,
-        Optional('epoch'): Any(int, None),
+        Optional('epoch'): Any(Coerce(int), None),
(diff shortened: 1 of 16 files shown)


curator/validators/filters.py
```diff
@@ -67,7 +67,7 @@ def prune_nones(mydict):
                 '{0}, filter #{1}: {2}'.format(location, idx, pruned)
             ).result()
             logger.debug('Filter #{0}: {1}'.format(idx, filter_dict))
-            v[idx] = pruned
+            v[idx] = filter_dict
```
untergeek: This didn't matter until the `Coerce(int)` was necessary.

PR #756, opened 2016-09-08
Make the envvars example more explicit

JD557:
The last example was not that helpful (setting a value to the default value).
Updated to make the evaluation behavior more explicit.


untergeek:
@JD557 can you [sign our CLA](http://www.elasticsearch.org/contributor-agreement/) please?  I can't merge even a doc change without that.

JD557:
Done.

I'm not sure if I need to do anything else to retry the check.

untergeek:
I manually verified.  Thanks for signing. LGTM™.

PR #786, opened 2016-10-19
Add missing `source` to 2nd age filter

jjm:
When using this example it produces a trackback due to `source` being missing from the younger age filter.


jjm:
I've completed the CLA  now, created the pull request too early :-(.

untergeek:
Thanks for catching that! I did so many examples at 4.0 release time that I must have overlooked that one.

jjm:
Your welcome. liking the way to configure 4.0 so much more than 3.x.

PR #803, opened 2016-11-07
Change indices in IndexList to be a deep copy of all_indices

DanilSerd:


karmi:
Hi @DanilSerd, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/curator/pull/803.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

untergeek:
You need to sign the CLA before I can merge this

PR #809, opened 2016-11-16
Fix es_repo_mgr path in setup.py

HonzaKral:
the settings in `setup.py` point to a non existent module causing all calls to `es_repo_mgr` to end with
```
ImportError: No module named es_repo_mgr
```


untergeek:
_Sigh._

This is what I get for migrating to precompiled stuff, you know?  I miss silly bits like this.

untergeek:
You'll probably have to wait for me to get back to the 'States to do a release with this.

I'll try to do it remotely, though.  The windows build will be the hard one...

PR #820, opened 2016-11-22
Release/4.2.3

untergeek:


untergeek:
Serious weirdness with this branch. It should be clean now, though.

PR #850, opened 2016-12-22
Fix version for dependency click

breml:


untergeek:
Thanks for correcting the typo.

PR #853, opened 2016-12-22
Make filter_list optional

alexef:
For `snapshot` and `show_snapshot` commands, we usually run them against the entire cluster.

Not sure about providing a string default, it can be fixed at a different level.

### DIFF
--- a/curator/singletons.py
+++ b/curator/singletons.py
@@ -473,7 +473,7 @@ def replicas_singleton(
 )
 @click.option(
     '--filter_list', callback=validate_filter_json,
-    help='JSON string representing an array of filters.', required=True
+    help='JSON string representing an array of filters.'
 )
 @click.pass_context
 def snapshot_singleton(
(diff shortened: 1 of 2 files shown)


untergeek:
I can't merge any code (even documentation changes) if you do not sign the [CLA](http://www.elasticsearch.org/contributor-agreement/)

alexef:
@untergeek I know, I signed, I'm waiting for the confirmation email :)

untergeek:
Also, simply removing `required=True` is not optimal.  I recommend replacing that with `default='{"filtertype":"none"}'`.

untergeek:
And since you're going to the trouble to do this, I would ask that you make that change to `show_indices` in addition to `snapshot` and `show_snapshot`.  Thanks.

untergeek:
And please rebase against the current master so you can properly add a line to the current Changelog, if you don't mind.

alexef:
I don't mind, updated.

Issue #861, opened 2017-01-03
verify_client_object


cp2587:
Hi,

Thank you for your work on this library. We recently updated to ES 5.0 and cannot use curator anymore because of the check that happens in 'verify_client_object', we subclassed elasticsearch.Elasticsearch and so the following line raise an exception:
```
elif not type(test) == type(elasticsearch.Elasticsearch()):  
     raise TypeError('Not a client object. Type: {0}'.format(type(test)))
```

Can i sugget you use something like this instead:
```
elif not isinstance(test, elasticsearch.Elasticsearch):  
     raise TypeError('Not a client object. Type: {0}'.format(type(test)))
```

untergeek:
Feel free to submit a PR. It has to pass all tests. You might need to change one of the unit tests accordingly.

cp2587:
Opened a PR: https://github.com/elastic/curator/pull/862 (i signed the CA after opening it, will it automatically update the checks ?)

untergeek:
Closed by #862

PR #862, opened 2017-01-04
use isinstance instead of type() in verify_client_object

cp2587:
Using isinstance is more robust and allow library user to subclass elasticsearch.Elasticsearch
Also add unittest for verify_client_object function
submitted following https://github.com/elastic/curator/issues/861


untergeek:
LGTM™

Thanks!

PR #864, opened 2017-01-09
Add information about the current package version

abraxxa:
The latest version of `python-elasticsearch-curator` in the Debian/Ubuntu repo is 4.1.2 which isn't compatible with ES 5.1.1.
I learned the hard way that it dies on startup without logging to the configured logfile because it isn't compatible and the ES cluster filled up the disk.

### DIFF
--- a/docs/asciidoc/installation.asciidoc
+++ b/docs/asciidoc/installation.asciidoc
@@ -219,7 +219,7 @@ enabled=1
 [[yum-binary]]
 === Binary Package Installation
 
-Starting with version 4.1.2, Curator comes with a binary version.  What this
+Starting with version 4.1.2, Curator is only distributed as binary version. If you have installed an older version you won't get any updates until you manually install the `elasticsearch-curator` package.  What this
 really means is that the source is compiled, and all required libraries are
 bundled with the `curator` binary, so there are no conflicts.
 


untergeek:
I've left a comment.  I cannot merge your changes until you've signed the CLA (yes, even for documentation changes), and make a few revisions.

untergeek:
@abraxxa I also note that the bottom of the YUM installation page has the following warning block:
 
>The elasticsearch-curator binary package cannot coexist with the older python-elasticsearch-curator packages. This collision is explicitly defined in packages since 4.1.2, but not in previous packages. If it is installed, you must uninstall the python-elasticsearch-curator package before installing the elasticsearch-curator binary package.

Perhaps reword accordingly, since there is a warning there with regard to the older python-elasticsearch-curator package.

untergeek:
I'm closing this as it's been a month without feedback or response.  In this time period, I've also added https://www.elastic.co/guide/en/elasticsearch/client/curator/current/version-compatibility.html and issue #874 was added, which seeks to address similar concerns differently.

abraxxa:
Our legal department has identified multiple, strong reasons why we can't sign your contributor agreement, like no liability limitations.
I won't contribute any patches in the future, please don't ask me to send a pull-request as fix for a reported issue.

PR #932, opened 2017-04-17
Added %G and %V to timestring filter to support ISO week numbers

IgorMiller:
Hi,

Here is my proposition to support ISO week numbers (in index names) by adding support for '%G' and '%V' to timestring in the filter. First, I convert the timestring to get the week number in Gregorian calendar, and then I fix it to be in ISO week number.
Python has these directives built-in only since 3.6.
It was tested and works in Python 2.7. I also added few tests to demonstrate the newly supported behaviour.

Please review this PR.

Thanks


untergeek:
This looks great!  Thanks for contributing to Curator.  This ought to work well with the new `period` filter!

PR #934, opened 2017-04-17
Fixed ISO week parsing code in utils

IgorMiller:
Related to https://github.com/elastic/curator/pull/932

I realized I had a mistake in my code. When comparing two strings, the result was always false for 9 first weeks in the year, because I was not adding leading 0's to the week number, and behaviour was incorrect. I added new tests to demonstrate it.

I had also to remove the comparison to support the case of the ISO leap year (53 weeks, but 52 in Gregorian). Ex:2009 week 53.
```
>>> week53 = datetime.strptime("2009.W531", "%Y.W%W%w")
>>> week53
datetime.datetime(2010, 1, 4, 0, 0)
# Converting back to Gregorian gives first week
>>> datetime.strftime(week53, "%Y.W%W%w")
'2010.W011'
# ISO calendar gives also first week. The comparison will return false (they are equal) and not adjust the date.
>>> week53.isocalendar()
(2010, 1, 1)
```
The above example was working because the iso_week_str was missing a leading 0. In conclusion, it is not because iso week number and gregorian week number are the same means that they are correct! We have always to compare gregorian string with the original one.

Sorry for the mistake. Feel free to add any other tests!

Thanks :)

### DIFF
--- a/curator/utils.py
+++ b/curator/utils.py
@@ -252,22 +252,19 @@ def fix_epoch(epoch):
 
 def _handle_iso_week_number(date, timestring, index_timestamp):
     date_iso = date.isocalendar()
-    iso_week_str = "{Y}{W}".format(Y=date_iso[0], W=date_iso[1])
+    iso_week_str = "{Y:04d}{W:02d}".format(Y=date_iso[0], W=date_iso[1])
     greg_week_str = datetime.strftime(date, "%Y%W")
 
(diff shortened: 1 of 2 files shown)


untergeek:
Will you please squash these commits?  It makes it much easier to keep the master and 5.x branches in sync.

Other than that, LGTM™

PR #990, opened 2017-06-14
Properly catch wildcard index restores

### LINKED ISSUE #989: Snapshot recovery must work with wildcards
https://discuss.elastic.co/t/curator-throwing-error-while-restoring-failed-to-complete-action/89308/11

The issue is [here](https://github.com/elastic/curator/blob/v5.1.1/curator/utils.py#L1447):
```
    for index in index_list:
```

An index named for a wildcard will never be in the `response`, as it's not an actual index name.

The proposed solution will be:

```
   for index in response:
```

and to increase logging so that the provided indices and the found indices are both adequately described.  This should help when a wildcard does not make any matches.  A message that the restore completed may occur, but the logs will reveal why it said that, now.

untergeek:
fixes #989


raj-krrish:
I am still facing this issue. Restore throws error while recovery check, but runs smoothly when I set wait_for_completion flag to False. I am using elasticsearch-curator (5.1.1)

Issue #996, opened 2017-07-04
Add a feature to pull unit_count for filters from index name


soenkeliebau:
We are currently looking into building a flexible log management architecture that gives the log sender some control over how his log events are treated by processing metadata about that log source.
One of those fields is a time to live in days, so how long that event should be kept, before it is deleted.

Events are then stored in indices called logstash-30-yyyy.mm.dd (for a 30 day ttl) or logstash-60-yyyy.mm.dd (for a 60 day ttl) etc.
In order to enable housekeeping on these we'd need to create an action in curator per existing value for ttl (unless I am missing something), which is not desireable.
Would it be of interest to add functionality to specify a regex or something similar to enable curator to dynamically pull the value for unit_count from the index name? I'd be happy to look into this and come up with a proposal of how configuration could look first and work on a PR later on if people think this makes sense.

untergeek:
It's a fascinating idea.  I cannot think that this could hope to work without a regular expression with a captured group, though.  That's certainly not a deal breaker, but it would be daunting for those not familiar with regular expressions.

The larger question for me is centered around the rollover API and action.  Elasticsearch is headed towards recommending _away_ from named indices in favor of using the rollover API.  While you clearly could name indices with a numeric value and operate in a similar way, it becomes less important to do so, as you end up rotating based on document count and/or age relative to creation_date, rather than index name.  Does this approach still offer value if you're doing that?  It could, I guess, but with such a rotation, it seems it would be more valuable to use the field_stats API (or the new API replacing it) to determine the age of the oldest and newest documents, so as to not delete anything not _fully_ out of the retention period.

soenkeliebau:
I agree with you on the regex that would be necessary for this to work. While in theory we could probably come up with a defined marker for the unit count and a syntax similar to the index name pattern the longer I thought about it the more complex I had to make it to accomodate the cases I came up with. Regexes have been around for a long time, are extremely powerful and well documented (as in there is a lof of documentation, not necessarily saying all of it is good) - so no need to reinvent the wheel.
I won't pretend to know the background of the average curator user, but I think a fair few of them have grep-ed some log files back in the day and used regular expressions to do so, so I am confident that this should work out nicely.

On your second question, I think that this will play quite nicely with the rollover feature, since the underlying principle is the same.
For example, if we have two different retention times, 30 and 60 days, if I don't misunderstand the rollover pattern, the indices for this might look like this:

logs-30-1
logs-30-2
logs-30-3
logs-60-1
[...]

PR #997, opened 2017-07-06
#996: Added functionality to dynamically filter indices based on …

soenkeliebau:
…their age and a value that is taken from the index name with a regex.

### DIFF
--- a/curator/defaults/filter_elements.py
+++ b/curator/defaults/filter_elements.py
@@ -122,6 +122,12 @@ def unit_count(**kwargs):
     # filtertype if use_age is set to True.
     return { Required('unit_count'): Coerce(int) }
 
+def unit_count_pattern(**kwargs):
+    # This setting is used with the age filtertype to define, whether
+    # the unit_count value is taken from the configuration or read from
+    # the index name via a regular expression
+    return { Optional('unit_count_pattern'): Any(str, unicode) }
+
 def use_age(**kwargs):
     # Use of this setting requires the additional setting, source.
     return { Optional('use_age', default=False): Any(bool, All(Any(str, unicode), Boolean())) }
--- a/curator/defaults/filtertypes.py
+++ b/curator/defaults/filtertypes.py
@@ -48,6 +48,7 @@ def age(action, config):
         filter_elements.direction(),
         filter_elements.unit(),
         filter_elements.unit_count(),
+        filter_elements.unit_count_pattern(),
         filter_elements.epoch(),
         filter_elements.exclude(),
     ]
--- a/curator/defaults/settings.py
+++ b/curator/defaults/settings.py
@@ -118,6 +118,7 @@ def structural_filter_elements():
         Optional('timestring'): Any(str, unicode, None),
         Optional('unit'): Any(str, unicode),
         Optional('unit_count'): Coerce(int),
+        Optional('unit_count_pattern'): Any(str, unicode),
         Optional('use_age'): Boolean(),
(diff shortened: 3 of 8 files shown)


untergeek:
I won't be able to address this PR this week, unfortunately.  I'll be attending (and presenting at) OpenWest.  I'll get to this as soon as I otherwise can.

soenkeliebau:
No worries, there really is no rush on this.
I'll see if I find some time to look at existing tests and try to come up with a few cases to test this functionality in the meantime.

untergeek:
Sorry for the delay.  I started a review.  It looks good!  Only cosmetic issues, plus adding the API documentation.

All we need now are some tests, and we'll be good to go.

untergeek:

soenkeliebau:
I've rebased on master and added some integration tests for the filtering functionality as well as fixed two small things regarding error handling, when an illegal regex is given or it contains no capture group.
Do we also want to add unit tests? I was unsure because I'd have to add new cluster and index settings to the testvars and it seemed like duplication of what we already test with the integration tests to me. 
But I am of course happy to do that as well.

untergeek:
If integration tests are handling it all, unit tests are redundant.  I tend to try to test as much as I can in unit tests, and then do some quick end-to-end tests, both positive and negative, in integration tests.  If you've added them all to integration, that's fine too, though.

I'll take another look and merge when done.

untergeek:
Got burned by this test:
```
    def test_retention_from_name_illegal_regex_with_fallback(self):
        # Test extraction of unit_count from index name when pattern contains an illegal regular expression
        # Create indices for 10 months with retention time of 2 months in index name
        # Expected: Fallback value of 3 is used and 3 most recent indices remain in place
        self.args['prefix'] = 'logstash_2_'
        self.args['time_unit'] = 'months'
        self.create_indices(10)
        self.write_config(
            self.args['configfile'], testvars.client_config.format(host, port))
        self.write_config(self.args['actionfile'],
                          testvars.delete_pattern_proto.format(
                              'age', 'name', 'older', '\'%Y.%m\'', 'months', 3, '_[0-9+_', ' ', ' ', ' '
                          )
                          )
        test = clicktest.CliRunner()
        result = test.invoke(
            curator.cli,
            [
                '--config', self.args['configfile'],
                self.args['actionfile']
            ],
        )
        self.assertEquals(3, len(curator.get_indices(self.client)))
```

Works great, except that 3 months is not calculated as calendar months, but 3*30 days, which means that running this test today (at this time) will fail, as 90 days deletes the 3rd index (July), as UTC rollover is older than 6am on the 1st:

```
DEBUG      curator.indexlist          filter_by_age:457  Unit_count_pattern is set, trying to match pattern to index "logstash_2_2017.07"
DEBUG      curator.indexlist          filter_by_age:472  Unable to match pattern using fallback value of "3"
DEBUG      curator.indexlist           __actionable:35   Index logstash_2_2017.07 is actionable and remains in the list.
DEBUG      curator.indexlist            __excludify:58   Remains in actionable list: Index "logstash_2_2017.07" age (1498867200), direction: "older", point of reference, (1498932790)
```
Going to have to convert this test to a unit test with a fixed `epoch` as a result.

untergeek:
Or...

Just not test months, but rather days, which don't suffer from that problem. 😄

soenkeliebau:
thanks for catching and fixing this @untergeek !

Do you want me to add a test for months with fixed epoch also?

untergeek:
I think the other tests are okay. Thanks, though!

PR #1010, opened 2017-07-27
Allow dates in new_index value for rollover action

tschroeder-zendesk:
Allow usage of date variables in the new_index option for the rollover action.

### DIFF
--- a/curator/actions.py
+++ b/curator/actions.py
@@ -863,7 +863,7 @@ class Rollover(object):
         self.settings   = extra_settings
         #: Instance variable.
         #: Internal reference to `new_index`
-        self.new_index = new_index
+        self.new_index = parse_date_pattern(new_index) if new_index else new_index
         #: Instance variable.
         #: Internal reference to `wait_for_active_shards`
         self.wait_for_active_shards = wait_for_active_shards
--- a/test/integration/test_rollover.py
+++ b/test/integration/test_rollover.py
@@ -310,3 +310,53 @@ class TestCLIRollover(CuratorTestCase):
                     ],
                     )
         self.assertEqual(expected, self.client.indices.get_alias(name=alias))
+    def test_max_age_with_new_name_with_date(self):
+        oldindex  = 'rolltome-000001'
+        newindex  = 'crazy_test-%Y.%m.%d'
+        alias     = 'delamitri'
+        condition = 'max_age'
+        value     = '1s'
+        expected  = {curator.parse_date_pattern(newindex): {u'aliases': {alias: {}}}}
+        self.client.indices.create(
+            index=oldindex,
+            body={ 'aliases': { alias: {} } }
+        )
+        time.sleep(1)
+        self.write_config(
+            self.args['configfile'], testvars.client_config.format(host, port))
+        self.write_config(self.args['actionfile'],
+            testvars.rollover_with_name.format(alias, condition, value, newindex))
+        test = clicktest.CliRunner()
+        result = test.invoke(
+                    curator.cli,
+                    [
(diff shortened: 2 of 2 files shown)


untergeek:
Thanks for contributing to Curator!  I have a concern. Have you tried to rollover an index named `crazy_test-%Y.%m.%d`?  

Elasticsearch does not allow you to rollover an index not ending with a dash and a number, e.g. `-1`, even with named indices.  The test you provided may work, but the problem would arise when you tried to rollover the resulting index named `crazy_test-%Y.%m.%d`.

See [here](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/indices-rollover-index.html#_using_date_math_with_the_rollover_api):

> It can be useful to use date math to name the rollover index according to the date that the index rolled over, e.g. logstash-2016.02.03. The rollover API supports date math, but requires the index name to end with a dash followed by a number, e.g. logstash-2016.02.03-1 which is incremented every time the index is rolled over.

The emphasis should be on the phrase, **but requires the index name to end with a dash followed by a number.**  It may work if your date string uses hyphens/dashes, but the reality is that if you _do_ use that date format, then testing for this condition quickly gets messy and ugly.

This patch can still be merged, but a check to see if the index name ends with a dash followed by a number would have to be added, along with unit tests (positive and negative), and hopefully integration tests also verifying the same behavior.

tschroeder-zendesk:
Yeah I'm adding tests now. I will make sure it works. Thanks.

tschroeder-zendesk:
This will work for anyone who wants to do a daily rollover like this:
 name-2017.07.26
 name-2017.07.27
 etc.

And that use case will work with this code. The case I want is really to be able to take in a bad named index and start rollover on it and then continue rollover with the same curator script which I can do but will need more work so I am going to do that as a different pull request. This one adds a feature and works fine. It's not amazing, but it doesn't introduce any new issues. Using new_index avoids the requirement of -#, but even before my change you could do new_index=blah and it would fail on the next run since blah would already exist. Anyways as such short of any more needed testing you all think I should add I am calling this done.

untergeek:
The Elasticsearch API is letting you do something that the documentation says it should not allow you to do.  I will be raising this with the Elasticsearch developers.  I dare not merge this functionality if this is a bug they plan on fixing, or a loophole they plan on closing.

tschroeder-zendesk:
I'm not sure why you think this is a bug. You are allowed to rollover any index if you provide a new_index name. It's in the documentation right here: 

https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-rollover-index.html#_naming_the_new_index

untergeek:
Because of the line in the date math section, referred to above.

I did get this clarified from the ES team, and so long as no users _try_ to use [date math](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/date-math-index-names.html) in the new name, then you can name each index whatever you want.  If, however, they want to use date math, then they must end with a dash and a number.

To verify that this works as expected, I would like to see 2 tests added that use a plain-text date math example (you don't need to do the escaping, it's handled automatically): one that passes, and one that fails because of the absence of a dash and a number at the end of the date-math-index-name.

Bonus points for a screwy new index name that combines strings for `parse_date_pattern` _and_ date math (just not nested, as I'm reasonably certain that will fail).

untergeek:

tschroeder-zendesk:
I ended up working around this issue. I will try to get to it to add more testing if I have time at the moment I need to put this on hold.

untergeek:
I may merge this, and add more tests myself later.  It's functional as it is.  I would like to add this to the next release.

Issue #987, opened 2017-06-11
Snapshots do not work with AWS ES 5.3


cjuroz:
As requested in #639 I open this issue:

AWS ES 5.3 has been released recently and they _now_ claim is compatible with Curator: https://aws.amazon.com/about-aws/whats-new/2017/06/elasticsearch-5-3-now-available-on-amazon-elasticsearch-service/
> Curator Support: Enabled usage of the popular Curator utility to simplify index management tasks.

I have been using Curator 5 API to create a Lambda function (here: https://gist.github.com/cjuroz/d45f4d73e74f068892c5e4f3d1c7fa7c) that will help to automate index cleaning and snapshots management. 

So far, index cleaning operations are working pretty fine *BUT* snapshot creation is failing because AWS guys still didn't expose `/_snapshot/_status` endpoint. I already reported this bug here: https://forums.aws.amazon.com/thread.jspa?threadID=257549.

Note: As a workaround, I am using ElasticSearch API to directly send signed request to ES to do snapshots.

untergeek:
I addressed this in the documentation in https://github.com/elastic/curator/pull/988

mkubenka:
It looks like `/_snapshot/_status` endpoint is working after AWS ES [add support for 5.5](https://aws.amazon.com/about-aws/whats-new/2017/09/elasticsearch-5_5-now-available-on-amazon-elasticsearch-service/). Probably related with availalibity to retore automated snapshots:

> Self-service restore of Amazon Elasticsearch Service domains is now available from the automated snapshots taken by the service. All automated snapshots are available for restore from the Elasticsearch API or Curator CLI. You no longer have to request support assistance for restore requests.

I have tried `examples/actions/snapshot.yml` and it worked even on 5.3!

untergeek:
From what I understand, this has been corrected in AWS ES 5.5.  Closing for now.

PR #1012, opened 2017-07-28
AWS IAM authentication and Snapshot creation

kobuskc:
Instead of hard coding AWS credentials, this will enable the user to enable request signing in the config.yml file and specify a region. The credentials will then be collected from the environment variables, local config file in ~/.aws/credentials or an instance profile if attached.

When making a snapshot where the cluster is running on AWS, a routine was added to still be able to check if any current snapshots are running before starting a new one. (issue #987 )


curator/utils.py
```diff
@@ -1159,7 +1156,27 @@ def snapshot_running(client):
     :rtype: bool
     """
     try:
-        status = client.snapshot.status()['snapshots']
+        if aws_flag:
```
untergeek: How does this function get the `aws_flag` variable?  It won't keep state inside the `utils.py` file, as that's just a reference library.  That's why the tests are not passing:
```
ERROR            curator.cli                    cli:193  Failed to complete action: snapshot.  <class 'curator.exceptions.FailedExecution'>: Exception encountered.  Rerun with loglevel DEBUG and/or check Elasticsearch logs for more information. Exception: global name 'aws_flag' is not defined
```

karmi:
Hi @kobuskc, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/curator/pull/1012.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

kobuskc:
Hi @karmi, I'm new to this. The e-mail address in my Github profile and the one I used in my commit are the same. It is the only one I use. In my Github profile, it only shows my address though, not my full name as well. Will that make a difference?

kobuskc:
Hi @untergeek , I have moved away from trying to keep `aws_flag` as a global variable and put it in `settings.py` instead.

kobuskc:
If this PR is accepted, request signing can be configured by setting adding the following to `curator.yml`:

`aws_sign_request: True
  aws_region: 'your-aws-region'`

The credentials required to sign the requests will then be retrieved from the environment by checking these locations:
- Environment variables
- AWS credentials file
- Instance profile credentials

If `aws_sign_request` is set to True, the `aws_flag` variable will be set, allowing the snapshot function to still check for current running snapshots before attempting a new one.

untergeek:
I'm not convinced the approach of putting `aws_flag` in `settings.py` is going to work reliably.  It's always being assigned as `False` in that file.  Any other file that opens it is going to potentially reset it as `False`.  A more consistent and deterministic approach should be used.

I presume that there are plugins installed, or other features, that identify an AWS ES node.  Start probing the results of `client.nodes.info()`:

```
In [10]: client.nodes.info()['nodes']['JLS3tYelSKy2XnK1ZOc3WA']['plugins']
Out[10]:
[{u'classname': u'org.elasticsearch.xpack.XPackPlugin',
  u'description': u'Elasticsearch Expanded Pack Plugin',
  u'has_native_controller': True,
  u'name': u'x-pack',
  u'version': u'5.4.3'}]
```
The keys at the node level that may be of benefit are:
```
[u'plugins',
 u'thread_pool',
 u'transport_address',
 u'http',
 u'name',
 u'roles',
 u'total_indexing_buffer',
 u'process',
 u'ip',
 u'settings',
 u'modules',
 u'ingest',
 u'host',
 u'version',
 u'jvm',
 u'build_hash',
 u'attributes',
 u'os',
 u'transport']
```
It is relatively easy to crawl through this list and find the relevant information.  Something here has to inform us that this is an AWS ES node.  

Feel free to paste the output in different boxes here (not one massive one, unless it's as a zip file).  I can compare it to a non-AWS ES 5.3 instance and see where the differences may be.

untergeek:
It could be something in the `os` section, for example:
```
client.nodes.info()['nodes']['JLS3tYelSKy2XnK1ZOc3WA']['os']
Out[17]:
{u'allocated_processors': 1,
 u'arch': u'amd64',
 u'available_processors': 1,
 u'name': u'Linux',
 u'refresh_interval_in_millis': 1000,
 u'version': u'4.10.15-1-pve'}
```

What version does an AWS ES instance show?

untergeek:
And perhaps, most importantly, the `build_hash`:
```
client.nodes.info()['nodes']['JLS3tYelSKy2XnK1ZOc3WA']['build_hash']
Out[23]: u'eed30a8'
```
By combining this with the output of `version` we can categorically identify an AWS ES node as separate from a non-AWS ES node, because AWS ES modifies the code from the ES release default.  

The known AWS ES version/build_hash map could be kept in the `settings.py` file, as it will not change.  When future releases come out, this list could be updated.  Of course, this hinges on the belief that AWS ES only has 1 version of the 5.3 code in deployment, so there would only be one build_hash in the wild.  If there turned out to be more, some way of specifying the build_hash or providing an override file would be a potential solution.

untergeek:
If we're going that route, just use `client.info()`:
```
In [25]: client.info()
Out[25]:
{u'cluster_name': u'untergeek',
 u'cluster_uuid': u'7OfiDNTsTJqiHTK4_ghIsA',
 u'name': u'esclient',
 u'tagline': u'You Know, for Search',
 u'version': {u'build_date': u'2017-06-22T00:34:03.743Z',
  u'build_hash': u'eed30a8',
  u'build_snapshot': False,
  u'lucene_version': u'6.5.1',
  u'number': u'5.4.3'}}
```
`version` is a hash/dict here that includes the `build_date`, the `build_hash`, and the release `number`.  Those together constitute enough of a map to differentiate between release ES and AWS ES.

karmi:
Hi @kobuskc, now the e-mail in the Git commit is the same as in the CLA signature, thanks for whatever you did to fix that! :)

kobuskc:
The version information doesn't include a proper build hash. I have checked on multiple clusters and it is always "Unknown"

I can also see that there is only limited information available in the os key, but that could help. I can check for the absence of os.name, os.arch and os.version and the presence of the discovery-ec2 plugin. 

Here's an example output:

```
client.nodes.info()['nodes']['FLzHlSuSQkmfP4to9MJR0A']['os']

{'allocated_processors': 2,
 'available_processors': 2,
 'refresh_interval_in_millis': 1000}

```
```
client.info()

{'cluster_name': '733061676469:helixsearch',
 'cluster_uuid': 'tFl0cnoQSVSOuOmheMswZw',
 'name': 'LmYkw-v',
 'tagline': 'You Know, for Search',
 'version': {'build_date': '2017-04-28T21:44:52.237Z',
  'build_hash': 'Unknown',
  'build_snapshot': False,
  'lucene_version': '6.4.2',
  'number': '5.3.2'}}
```
Please let me know if this approach is acceptable.

untergeek:
With AWS ES being tied (currently) to v5.3.2 (are there other minor versions for AWS ES?), you can probably check for a combination of 5.3.whatever, a build_hash value of `Unknown`, and the presence of the `discovery-ec2` plugin to safely assume an AWS ES instance.  We can include other versions when AWS ES adds them.  

A variable in `settings.py`, as an array of version tuples, like `aws_es_versions = [ (5,3,2) ]` would probably be the best place to store this.  An example of version comparison can be seen in the [`get_version`](https://github.com/elastic/curator/blob/v5.1.1/curator/utils.py#L568-L582) function in `utils.py`.

untergeek:

curator/utils.py
```diff
@@ -1173,9 +1173,9 @@ def snapshot_running(client):
     :rtype: bool
     """
     try:
-        if settings.aws_flag:
+        # Check to see if we're running AWS Elasticsearch and if so, iterate through each repository to check for running snapshots
+        if settings.aws_versions['5.3.2'] == client.info()['version']:
```
untergeek: Maybe a loop here?

```python
is_aws = False
found_version = client.info()['version']
for aws_version in settings.aws_versions:
    if found_version == aws_version:
        is_aws = True
```
That way it will set the flag true if even one of the AWS versions matches.  It won't matter if the others don't.

untergeek: Then you'll have a valid boolean with `is_aws`

kobuskc: Thanks. That makes a lot of sense. I have made the required changes.

untergeek:
There are 4 tests that are not passing.  Each is a unit test, and each is failing due to the new line added to the `snapshot_running` function in utils.py.

```
found_version = client.info()['version']
```

This call needs to be properly mocked in these four tests for the tests to come back passing.

The tests are:

1. `ERROR: test_do_action_snap_in_progress (test.unit.test_action_restore.TestActionRestore)`
1. `ERROR: test_do_action_success_no_wfc (test.unit.test_action_restore.TestActionRestore)`
1. `ERROR: test_false (test.unit.test_utils.TestSnapshotRunning)`
1. `ERROR: test_true (test.unit.test_utils.TestSnapshotRunning)`

You should add a line to each of these tests like this:

```
client.into.return_value = testvars.es_release_version
```
In testvars.py, you should add an entry like this:

```
es_release_version = { 'version': {u'build_date': u'2017-06-22T00:34:03.743Z', u'build_hash': u'eed30a8', u'build_snapshot': False, u'lucene_version': u'6.5.1', u'number': u'5.4.3'} }
```
This will at least get those tests to pass.

Since there is no way to properly test against an AWS ES version, it would be good if you also made new tests that were "clones" of these same 4 tests, but that use an AWS ES version, so that they test the `if is_aws` functionality:

```
aws_release_version = { 'version': {'build_date': '2017-04-28T21:44:52.237Z','build_hash': 'Unknown','build_snapshot': False,'lucene_version': '6.4.2','number': '5.3.2'} }
```
It will be more work to properly mock those calls, but it is the best way to ensure this works across multiple python versions.

kobuskc:
I have updated the tests that broke due to the new line I had added earlier. I will add AWS specific tests as soon as possible.

untergeek:
@kobuskc You probably need to do some rebasing before this can be committed.  Also, since AWS ES has a 5.5 version, you may need to include that in your hash/array.

untergeek:
Any update on this @kobuskc ?

kobuskc:
@untergeek, sorry it took so long. I have added the unit tests, new AWS ES 5.5 version and rebased.

curator/utils.py
```diff
@@ -748,27 +745,25 @@ def get_client(**kwargs):
 
     try:
         from requests_aws4auth import AWS4Auth
-        kwargs['aws_key'] = False if not 'aws_key' in kwargs \
```
untergeek: Question.  If a user is currently using `aws_key` and `aws_secret_key`, won't this result in a breaking change to them if they upgrade?  I'm concerned about breaking someone's existing functionality in a minor version change.

This would be an easy change for a major revision, but if there's no way to preserve existing functionality, that's going to be a problem.

curator/utils.py
```diff
@@ -1189,8 +1184,26 @@ def snapshot_running(client):
     :arg client: An :class:`elasticsearch.Elasticsearch` client object
     :rtype: bool
     """
+    status = []
+
     try:
-        status = client.snapshot.status()['snapshots']
+        # Check to see if we're running AWS Elasticsearch and if so, iterate through each repository to check for running snapshots
```
mkubenka: Is this still necessary? [`/_snapshot/_status` endpoint is working after AWS ES add support for 5.5](https://github.com/elastic/curator/issues/987#issuecomment-333255862)

curator/utils.py
```diff
@@ -748,27 +745,25 @@ def get_client(**kwargs):
...
-        kwargs['aws_region'] = False if not 'aws_region' in kwargs \
-            else kwargs['aws_region']
-        if kwargs['aws_key'] or kwargs['aws_secret_key'] or kwargs['aws_region']:
-            if not kwargs['aws_key'] and kwargs['aws_secret_key'] \
-                    and kwargs['aws_region']:
-                raise MissingArgument(
-                    'Missing one or more of "aws_key", "aws_secret_key", '
-                    'or "aws_region".'
-                )
+        from boto3 import session
```
mkubenka: Missing error message if boto3 is not present.

mkubenka:
Also, please don't forget link to boto3 [Credentials](http://boto3.readthedocs.io/en/latest/guide/configuration.html) manual page in documentation for all possible ways of setting credentials.

curator/utils.py
kobuskc: I have tested the original code agains an AWS ES 5.3 and 5.5 cluster and it works as expected now, so this is not necessary anymore.

curator/utils.py
kobuskc: Thanks. I have added a catch to print an error message if the import fails and exit.

curator/utils.py
kobuskc: This makes sense. I have added an update to allow for backward compatibility for any existing functionality in a minor version change.

kobuskc:
@untergeek since the AWS ES 5.x releases now support /_snapshot/_status, I suggest we close this pull request and I create a new one with just the authentication updates. Do you agree?

curator/utils.py
untergeek: I'm not seeing it yet.  Where is it?  It should still allow for these args you've removed.  It has to be there, in case users are running from a non-AWS location (hey, it could happen).

curator/utils.py
```diff
@@ -748,27 +745,25 @@ def get_client(**kwargs):
...
+        kwargs['aws_sign_request'] = False if not 'aws_sign_request' in kwargs \
+            else kwargs['aws_sign_request']
+        if kwargs['aws_sign_request']:
+            settings.aws_flag = True
+            session = session.Session()
+            credentials = session.get_credentials()
```
untergeek: I'm not seeing the reverse compatibility here yet.  Perhaps a `try/except` block that falls back to the old credentials?

untergeek:
@kobuskc It's not a bad idea to do that. Feel free to do so.

curator/utils.py
mkubenka: Actually, this PR is making credentials handling more standardised because it's delegated to *boto3* `Session`. There are [many ways](http://boto3.readthedocs.io/en/latest/guide/configuration.html) how user can configure credentials - environment variables, profiles, STS credentials,...

But, I definitely agree there should be backward compatibility and maybe deprecate options with next major release.

kobuskc:
New pull request: #1084

PR #1063, opened 2017-09-20
format error correctly

sevoma:
replace {0} with self.shrink_node

Previously, logging looked like this, because the string was never formatted:
```2017-09-20 16:51:15,396 INFO      Node "{0}" has multiple data paths and will not be used for shrink operations.```

### DIFF
--- a/curator/actions.py
+++ b/curator/actions.py
@@ -1882,6 +1882,7 @@ class Shrink(object):
             raise ActionError(
                 'Node "{0}" has multiple data paths and cannot be used '
                 'for shrink operations.'
+                .format(self.shrink_node)
             )
         self.shrink_node_avail = (
             self.client.nodes.stats()['nodes'][node_id]['fs']['total']['available_in_bytes']

### CODE REVIEW

File: curator/actions.py [hunk 1]
```diff
@@ -1880,7 +1880,7 @@ def qualify_single_node(self):
             raise ActionError('Node "{0}" is not usable as a shrink node'.format(self.shrink_node))
         if not single_data_path(self.client, node_id):
             raise ActionError(
-                'Node "{0}" has multiple data paths and cannot be used '
+                'Node "{0}" has multiple data paths and cannot be used '.format(self.shrink_node) + \
```
untergeek: Please put it on the last line before the closing parenthesis.  You don't need to use the `+ \` syntax that way, and it should work just fine.

untergeek: If not, I stand corrected.

untergeek: Confirmed that it will work that way (in `ipython`):
```
In [2]: print(
   ...:     'a-'
   ...:     'b-{0}-'
   ...:     'c'.format('FOO')
   ...: )
a-b-FOO-c
```

untergeek:
Thanks for catching this and submitting a pull request.  Please make the requested change, and I will be happy to merge this.

sevoma:
Done!

Issue #1066, opened 2017-09-24
[FEATURE] Add support for returning indices smaller than `disk_space` in space filter (PR enclosed)


dustin-decker:
Feature request:
Support returning indices *smaller* than `disk_space` in the space filter.

I needed a good way to use the shrink API on a bunch of different small indices (smaller than `disk_space` setting) that had too many shards. By adding a toggle for the comparison operator in the space filter, I was able to do this. I will add the pull request to this issue.

sevoma:
PR enclosed in https://github.com/elastic/curator/pull/1067

untergeek:
Closed by #1067

PR #1067, opened 2017-09-24
[FEATURE] Add support for returning indices smaller than `disk_space` in space filter

dustin-decker:
Addresses https://github.com/elastic/curator/issues/1066

I'm not attached to the `index_comparison` key at all, and hope there's a better name for it. I'm open to suggestions.
I'm not very familiar with asciidoc, so I hope I did that part right. If you want me to make any changes, just let me know.

Thanks a lot for Curator. It's a great tool, and I make heavy use of it.

### DIFF
--- a/curator/defaults/filter_elements.py
+++ b/curator/defaults/filter_elements.py
@@ -105,6 +105,10 @@ def timestring(**kwargs):
     else:
         return { Optional('timestring', default=None): Any(str, unicode, None) }
 
+def threshold_behavior(**kwargs):
+    # This setting is only used with the space filtertype and defaults to 'greater_than'.
+    return { Optional('threshold_behavior', default='greater_than'): Any('greater_than', 'less_than') }
+
 def unit(**kwargs):
     # This setting is only used with the age filtertype, or with the space
     # filtertype if use_age is set to True.
--- a/curator/defaults/filtertypes.py
+++ b/curator/defaults/filtertypes.py
@@ -118,6 +118,7 @@ def space(action, config):
         filter_elements.reverse(),
         filter_elements.use_age(),
         filter_elements.exclude(),
+        filter_elements.threshold_behavior(),
     ]
     retval += _age_elements(action, config)
     return retval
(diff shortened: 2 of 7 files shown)


untergeek:
Let me think about the name of the flag.  I like what you've done here, especially guaranteeing reverse-compatibility.  But this is a big shift in what the filter does.  I definitely want the name to be a good fit, and `index_comparison` doesn't quite seem the right fit.  I will think about this for a bit.

sevoma:
Maybe a bool? `smaller_than=True` (default False)

untergeek:
While I still consider the naming for this, will you please rebase against the master branch?  This commit has conflicts now.

dustin-decker:
Rebased

untergeek:
Thanks for rebasing.  

I like the idea of preserving the current behavior with a named approach, `threshold_behavior` as the key, with acceptable values of `less_than` and `greater_than`.  Of course, `greater_than` would be the default behavior to preserve reverse compatibility.

Thoughts?

dustin-decker:
I think that's a big improvement in clarity. What do you think of `disk_space_behavior` for the key? You already have the `disk_space` key, so it's slightly more obvious than 'threshold'. It also places it in close proximity to disk_space when alphabetically sorted, which is helpful for documentation.
I'll try to get the changes done, whatever the end up being, tomorrow or the day after.

untergeek:
I'm not a fan of mixing `disk_space` and `behavior` because the disk space isn't _behaving._  We have a threshold that is established by the value of `disk_space`, and then we need to do something with that value.  Here are some other possibilities:

* `threshold_calculation`
* `threshold_direction`
* `threshold_measurement`
* `threshold_comparison`

curator/defaults/filter_elements.py
```diff
@@ -105,6 +105,10 @@ def timestring(**kwargs):
     else:
         return { Optional('timestring', default=None): Any(str, unicode, None) }
 
+def threshold_behavior(**kwargs):
+    # This setting is only used with the space filtertype and defaults to 'greater_than'.
+    return { Optional('threshold_behavior', default='larger'): Any('greater_than', 'less_than') }
```
untergeek: This is going to cause a problem:

```
default='larger'
```

This has to be `greater_than` or we'll have serious issues.

dustin-decker: Good catch. I'm still reviewing my changes :)

dustin-decker: I should probably have a test for the default behavior

PR #1072, opened 2017-09-26
Add a verify option to es_repo_mgr create

alexef:
Sometimes (when we want to restore a backup) we don't want to verify the repository after creation.

Adding a flag to es_repo_mgr script, allowing us to disable the check, by passing `--verify false`.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -11,6 +11,11 @@ Changelog
     patterns.  This has been rectified, and an integration test added to 
     satisfy this.  Reported in #989 (untergeek)
 
+**General**
+
+  * The ``es_repo_mgr create`` command now can take ``verify`` as an argument
+    (default is True).
+
 5.1.1 (8 June 2017)
 
 **Bug Fixes**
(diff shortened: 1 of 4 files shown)


untergeek:
I'm out of town on a business trip this week.  I'll look it over more closely when I get back.

untergeek:
I like the idea, but can you match the name to what's used in Curator?  It's called [`skip_repo_fs_check`](https://www.elastic.co/guide/en/elasticsearch/client/curator/current/option_skip_fsck.html).  I realize that complicates your code, but it would be consistent with what's already in use.

untergeek:
This commit will need to be rebased on master, and the new feature added under the 5.3.0 part of the changelog.

alexef:
I'll take care of rebase and  consistency, and update this PR.

alexef:
Rebased against master, updated Changelog, changed flag name to skip_repo_fs_check. Was not able to test the latest changes.

untergeek:
You may need to rebase again, as I updated the Changelog recently.

alexef:
Updated.

PR #1082, opened 2017-10-12
Check to see if shards are on target node first

### LINKED ISSUE #1077: Shrink API should not move shards if they already exist on the same node
I have had issues with some failures on my side (bad config or other issues) and the result was that it would move the indices and then fail to assign shards for the shrink target. The state this leaves things in though is the shards are now all on one node ready to shrink. So, I re-run shrink and it goes oh this node has more space let me move one of each shard here. If possible add a check that sees if the condition of them all being on the same node is already met rather and skip the DETERMINISTIC node search. Just an idea I might look at implementing this myself if people agree with this tweak.

untergeek:
before re-issuing an allocation to them.

fixes #1077

### DIFF
--- a/curator/actions.py
+++ b/curator/actions.py
@@ -1999,16 +1999,20 @@ class Shrink(object):
             )
             self.__log_action(error_msg, dry_run)
 
-    def _check_all_shards(self, idx):
+
+    def _all_shards_on_box(self, idx):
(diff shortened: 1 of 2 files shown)


untergeek:
False alarm.  Code not needed.

PR #1083, opened 2017-10-12
test default case for threshold behavior

dustin-decker:
Title says it. Just adding the default case test for the bug you caught on #1067.

Thanks.


untergeek:
LGTM™

PR #1084, opened 2017-10-13
Add support for AWS IAM authentication using boto3

kobuskc:
Using the get_credentials definition, we can get credentials from the local environment.

Definition: http://boto3.readthedocs.io/en/latest/reference/core/session.html#boto3.session.Session.get_credentials

Order in which credentials is searched: http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#config-settings-and-precedence

### DIFF
--- a/curator/defaults/client_defaults.py
+++ b/curator/defaults/client_defaults.py
@@ -14,7 +14,9 @@ def config_client():
         Optional('client_key', default=None): Any(None, str, unicode),
         Optional('aws_key', default=None): Any(None, str, unicode),
         Optional('aws_secret_key', default=None): Any(None, str, unicode),
-        Optional('aws_region', default=None): Any(None, str, unicode),
+        Optional('aws_token', default=None): Any(None, str, unicode),
+        Optional('aws_sign_request', default=False): Boolean(),
+        Optional('aws_region', default='us-east-1'): Any(None, str, unicode),
         Optional('ssl_no_validate', default=False): Boolean(),
         Optional('http_auth', default=None): Any(None, str, unicode),
         Optional('timeout', default=30): All(
--- a/curator/defaults/client_defaults.py
+++ b/curator/defaults/client_defaults.py
@@ -16,7 +16,7 @@ def config_client():
         Optional('aws_secret_key', default=None): Any(None, str, unicode),
         Optional('aws_token', default=None): Any(None, str, unicode),
         Optional('aws_sign_request', default=False): Boolean(),
-        Optional('aws_region', default='us-east-1'): Any(None, str, unicode),
+        Optional('aws_region'): Any(None, str, unicode),
         Optional('ssl_no_validate', default=False): Boolean(),
         Optional('http_auth', default=None): Any(None, str, unicode),
         Optional('timeout', default=30): All(
--- a/curator/utils.py
+++ b/curator/utils.py
@@ -751,35 +751,47 @@ def get_client(**kwargs):
                     import certifi
                     kwargs['verify_certs'] = True
                     kwargs['ca_certs'] = certifi.where()
-    try:
-        from requests_aws4auth import AWS4Auth
-        kwargs['aws_sign_request'] = False if not 'aws_sign_request' in kwargs \
-            else kwargs['aws_sign_request']
-        if kwargs['aws_sign_request']:
-            try:
-                from boto3 import session
-            except ImportError as e:
-                logger.debug('Failed to import the "boto3" module.')
-                exit(e)
+    kwargs['aws_key'] = False if not 'aws_key' in kwargs \
+        else kwargs['aws_key']
+    kwargs['aws_secret_key'] = False if not 'aws_secret_key' in kwargs \
+        else kwargs['aws_secret_key']
+    kwargs['aws_token='] = '' if not 'aws_token' in kwargs \
+        else kwargs['aws_token']
+    kwargs['aws_sign_request'] = False if not 'aws_sign_request' in kwargs \
+        else kwargs['aws_sign_request']
+    kwargs['aws_region'] = False if not 'aws_region' in kwargs \
+        else kwargs['aws_region']
+    if not kwargs['aws_region']:
+        raise MissingArgument(
+            'Missing "aws_region".'
+        )
+    if kwargs['aws_sign_request']:
+        try:
+            from boto3 import session
+            from botocore import exceptions
+        # We cannot get credentials without the boto3 library, so we cannot continue
+        except ImportError as e:
+            logger.debug('Failed to import a module: %s' % e)
+            exit('Failed to import a module: %s' % e)
+        try:
             session = session.Session()
             credentials = session.get_credentials()
             kwargs['aws_key'] = credentials.access_key
             kwargs['aws_secret_key'] = credentials.secret_key
             kwargs['aws_token'] = credentials.token
-        else:
-            kwargs['aws_key'] = False if not 'aws_key' in kwargs \
-                else kwargs['aws_key']
-            kwargs['aws_secret_key'] = False if not 'aws_secret_key' in kwargs \
-                else kwargs['aws_secret_key']
-            kwargs['aws_token='] = '' if not 'aws_token' in kwargs \
-                else kwargs['aws_token']
+        # If an attribute doesn't exist, we were not able to retrieve credentials as expected so we can't continue
+        except AttributeError:
+            logger.debug('Unable to locate AWS credentials')
+            exit('Unable to locate AWS credentials')
+    try:
+        from requests_aws4auth import AWS4Auth
         if kwargs['aws_key'] or kwargs['aws_secret_key'] or kwargs['aws_region']:
             if not kwargs['aws_key'] and kwargs['aws_secret_key'] \
-                     and kwargs['aws_region']:
+                    and kwargs['aws_region']:
                 raise MissingArgument(
                     'Missing one or more of "aws_key", "aws_secret_key", '
                     'or "aws_region".'
-                    )
+                )
             # Override these kwargs
             kwargs['use_ssl'] = True
             kwargs['verify_certs'] = True
--- a/curator/utils.py
+++ b/curator/utils.py
@@ -647,11 +647,13 @@ def check_master(client, master_only=False):
 
 def get_client(**kwargs):
     """
-    NOTE: AWS IAM parameters `aws_key`, `aws_secret_key`, and `aws_region` are
-    provided for future compatibility, should AWS ES support the
-    ``/_cluster/state/metadata`` endpoint.  So long as this endpoint does not
-    function in AWS ES, the client will not be able to use
-    :class:`curator.indexlist.IndexList`, which is the backbone of Curator 4
+    NOTE: AWS IAM parameters `aws_sign_request` and `aws_region` are
+     provided to facilitate request signing. The credentials will be
+     fetched from the local environment as per the AWS documentation:
+     http://amzn.to/2fRCGCt
+
+    AWS IAM parameters `aws_key`, `aws_secret_key`, and `aws_region` are
+    provided for users that still have their keys included in the Curator config file.
 
     Return an :class:`elasticsearch.Elasticsearch` client object using the
     provided parameters. Any of the keyword arguments the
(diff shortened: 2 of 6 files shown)


curator/utils.py
```diff
@@ -745,36 +751,48 @@ def get_client(**kwargs):
...
         from requests_aws4auth import AWS4Auth
-        kwargs['aws_key'] = False if not 'aws_key' in kwargs \
-            else kwargs['aws_key']
-        kwargs['aws_secret_key'] = False if not 'aws_secret_key' in kwargs \
-            else kwargs['aws_secret_key']
-        kwargs['aws_region'] = False if not 'aws_region' in kwargs \
-            else kwargs['aws_region']
+        kwargs['aws_sign_request'] = False if not 'aws_sign_request' in kwargs \
+            else kwargs['aws_sign_request']
+        if kwargs['aws_sign_request']:
```
untergeek: This is going to need to be more complex.

Not only does it have to `try`/`except` on the import module, it has to `try`/`except` on the `session.Session()` (because it will raise an exception if it is unable, and we should control how it acts if it does, and log good messages).  If it is unable to get the session, then it would then rely on the original `aws_key` and `aws_secret_key`, which should be set first.

The contents of the `else` block should precede the `if kwargs['aws_sign_request']` statement as a pre-population thing, even if they end up being `False` because they were not specified.  The `boto3` stuff will overwrite the necessary values if it exists, so it is completely okay to have `kwargs['aws_key']` and `kwargs['aws_secret_key']` set first.

Does this make sense?

curator/defaults/client_defaults.py
```diff
@@ -14,7 +14,9 @@ def config_client():
         Optional('client_key', default=None): Any(None, str, unicode),
         Optional('aws_key', default=None): Any(None, str, unicode),
         Optional('aws_secret_key', default=None): Any(None, str, unicode),
-        Optional('aws_region', default=None): Any(None, str, unicode),
+        Optional('aws_token', default=None): Any(None, str, unicode),
+        Optional('aws_sign_request', default=False): Boolean(),
+        Optional('aws_region', default='us-east-1'): Any(None, str, unicode),
```
untergeek: Do you think it's wise to set the default to `us-east-1`, rather than just leave it kind of required to be set if you're using AWS credentials?

kobuskc: Makes sense. I have removed the default.

curator/utils.py
kobuskc: This makes sense. I have implemented your suggestions, so now credentials retrieved with boto3 will overwrite anything that was in the config file if the user specified both.

kobuskc:
Not sure why the Travis build failed. This is my test output:

----------------------------------------------------------------------
XML: /curator/nosetests.xml
Name                                  Stmts   Miss  Cover
---------------------------------------------------------
curator/__init__.py                      11      0   100%
curator/_version.py                       1      0   100%
curator/actions.py                      870    113    87%
curator/cli.py                          113      0   100%
curator/config_utils.py                  34      5    85%
curator/defaults/__init__.py              0      0   100%
curator/defaults/client_defaults.py       5      0   100%
curator/defaults/filter_elements.py      66      0   100%
curator/defaults/filtertypes.py          60      0   100%
curator/defaults/option_defaults.py     147      0   100%
curator/defaults/settings.py             32      0   100%
curator/exceptions.py                     9      0   100%
curator/indexlist.py                    369      0   100%
curator/logtools.py                      36      0   100%
curator/repomgrcli.py                    78      1    99%
curator/snapshotlist.py                 186      0   100%
curator/utils.py                        688     11    98%
curator/validators/__init__.py            1      0   100%
curator/validators/actions.py            19      0   100%
curator/validators/config_file.py         4      0   100%
curator/validators/filters.py            32      0   100%
curator/validators/options.py            13      0   100%
curator/validators/schemacheck.py        41      0   100%
---------------------------------------------------------
TOTAL                                  2815    130    95%
----------------------------------------------------------------------
Ran 570 tests in 136.390s

OK (SKIP=4)
Shutting down....
[root@xxxxxxxxxxx curator]# 
[nosetests.xml.gz](https://github.com/elastic/curator/files/1384222/nosetests.xml.gz)

curator/utils.py
```diff
@@ -745,15 +751,40 @@ def get_client(**kwargs):
...
+        try:
+            session = session.Session()
+            credentials = session.get_credentials()
+            kwargs['aws_key'] = credentials.access_key
+            kwargs['aws_secret_key'] = credentials.secret_key
+            kwargs['aws_token'] = credentials.token
+        # If an attribute doesn't exist, we were not able to retrieve credentials as expected so we can't continue
+        except AttributeError:
+            logger.debug('Unable to locate AWS credentials')
+            exit('Unable to locate AWS credentials')
```
untergeek: Don't do an `exit` here.  This should simply raise an Exception of some kind, either a `ConfigurationError`, indicating something is misconfigured, or a `CuratorException`, as there isn't an exception in exceptions.py that's more specific.

```
except AttributeError:
    raise CuratorException('Unable to locate AWS credentials, even though "aws_sign_request" was specified')
```

curator/utils.py
```diff
@@ -745,15 +751,40 @@ def get_client(**kwargs):
...
+        try:
+            session = session.Session()
+            credentials = session.get_credentials()
+            kwargs['aws_key'] = credentials.access_key
+            kwargs['aws_secret_key'] = credentials.secret_key
+            kwargs['aws_token'] = credentials.token
+        # If an attribute doesn't exist, we were not able to retrieve credentials as expected so we can't continue
+        except AttributeError:
+            logger.debug('Unable to locate AWS credentials')
+            raise exceptions.NoCredentialsError
```
untergeek: `NoCredentialsError` is not a valid member of the [base exceptions class](https://docs.python.org/3/library/exceptions.html).  Is that in the `boto` module?  If so, it would just be `NoCredentialsError`, rather than `exceptions.NoCredentialsError`.

untergeek:
Something is still not working with this code.  Client instances are not being created, causing every integration test to fail.

kobuskc:
Changed the import of botocore exceptions for clarity and fixed the logic error that prevented Client instances from being created. Local test results:

----------------------------------------------------------------------
XML: /curator/nosetests.xml
Name                                  Stmts   Miss  Cover
---------------------------------------------------------
curator/__init__.py                      11      0   100%
curator/_version.py                       1      0   100%
curator/actions.py                      870    113    87%
curator/cli.py                          113      0   100%
curator/config_utils.py                  34      5    85%
curator/defaults/__init__.py              0      0   100%
curator/defaults/client_defaults.py       5      0   100%
curator/defaults/filter_elements.py      66      0   100%
curator/defaults/filtertypes.py          60      0   100%
curator/defaults/option_defaults.py     147      0   100%
curator/defaults/settings.py             32      0   100%
curator/exceptions.py                     9      0   100%
curator/indexlist.py                    369      0   100%
curator/logtools.py                      36      0   100%
curator/repomgrcli.py                    78      1    99%
curator/snapshotlist.py                 186      0   100%
curator/utils.py                        710     29    96%
curator/validators/__init__.py            1      0   100%
curator/validators/actions.py            19      0   100%
curator/validators/config_file.py         4      0   100%
curator/validators/filters.py            32      0   100%
curator/validators/options.py            13      0   100%
curator/validators/schemacheck.py        41      0   100%
---------------------------------------------------------
TOTAL                                  2837    148    95%
----------------------------------------------------------------------
Ran 570 tests in 136.873s

OK (SKIP=4)
Shutting down....

untergeek:
So, is this connecting using `boto` now on your end?

If that part is working, the next step will be to add the proper documentation to [`docs/asciidoc/configuration.asciidoc`](https://github.com/elastic/curator/blob/v5.2.0/docs/asciidoc/configuration.asciidoc).  You may even need to edit existing information with regards to `aws_key` and the other related fields.  As these changes you are adding are not tested by the Travis CI testing, please leave them as such: hand tested.  I do not have the capability to test them, so they are a community addition, and should be properly vetted and tested by end users with the `--dry-run` flag, etc. before using.

kobuskc:
Yes, it connects with <boto> as expected. Here's an example test with <--dry-run>:

### With aws_sign_requests: True
xxxxxxxxxx  ~/actions  » curator --dry-run snapaction.yml 
2017-10-19 08:24:21,613 INFO      Preparing Action ID: 1, "snapshot"
**2017-10-19 08:24:21,756 INFO      Found credentials in shared credentials file: ~/.aws/credentials**
2017-10-19 08:24:21,817 INFO      Trying Action ID: 1, "snapshot": Snapshot members prefixed indices older than 1 day (based on index creation_date) with the default snapshot name pattern of 'curator-%Y%m%d%H%M%S'.  Wait for the snapshot to complete.  Do not skip the repository filesystem access check.  Use the other options to create the snapshot.
2017-10-19 08:24:21,875 INFO      DRY-RUN MODE.  No changes will be made.
2017-10-19 08:24:21,875 INFO      DRY-RUN: snapshot: curator-20171018212421 in repository backups with arguments: {'ignore_unavailable': False, 'partial': False, 'indices': 'members', 'include_global_state': True}
2017-10-19 08:24:21,875 INFO      Action ID: 1, "snapshot" completed.
2017-10-19 08:24:21,875 INFO      Job completed. 

### With no aws_sign_requests, and both aws_key and aws_secret_key specified
xxxxxxxxxx  ~/actions  » curator --dry-run snapaction.yml
2017-10-19 08:26:03,700 INFO      Preparing Action ID: 1, "snapshot"
2017-10-19 08:26:03,839 INFO      Trying Action ID: 1, "snapshot": Snapshot members prefixed indices older than 1 day (based on index creation_date) with the default snapshot name pattern of 'curator-%Y%m%d%H%M%S'.  Wait for the snapshot to complete.  Do not skip the repository filesystem access check.  Use the other options to create the snapshot.
2017-10-19 08:26:03,890 INFO      DRY-RUN MODE.  No changes will be made.
2017-10-19 08:26:03,890 INFO      DRY-RUN: snapshot: curator-20171018212603 in repository backups with arguments: {'partial': False, 'ignore_unavailable': False, 'include_global_state': True, 'indices': 'members'}
2017-10-19 08:26:03,890 INFO      Action ID: 1, "snapshot" completed.
2017-10-19 08:26:03,890 INFO      Job completed.

I will update the documentation today.

docs/asciidoc/configuration.asciidoc
```diff
@@ -420,6 +426,37 @@ IMPORTANT: You must set your <<hosts,hosts>> to the proper hostname _with_ port.
     It may not work setting <<port,port>> and <<hosts,hosts>> to only a host
     name due to the different connection module used.
 
+[[aws_sign_request]]
+=== aws_sign_request
+
+WARNING: This feature has not been fully tested and should be considered BETA.
+
+WARNING: This setting will not work unless the `requests-aws4auth` and `boto3` Python
+    modules have been manually installed first.
```
untergeek: This should probably read, "If installing via `pip`, this setting will not work unless ...".  I install `requests-aws4auth` in the [RPM/DEB](https://github.com/elastic/curator/blob/v5.2.0/unix_packages/build_official_package.sh#L143) packages.  If you want to edit both of the files in `unix_packages` directory to add `boto` to be installed, that would be okay, too.

docs/asciidoc/configuration.asciidoc
```diff
@@ -420,6 +426,37 @@ IMPORTANT: You must set your <<hosts,hosts>> to the proper hostname _with_ port.
...
+
+    . Environment variables
+    . Shared credential file (~/.aws/credentials)
+    . AWS config file (~/.aws/config)
+    . Boto2 config file (/etc/boto.cfg and ~/.boto)
+    . Instance metadata service on an Amazon EC2 instance that has an IAM role configured.
+
+[source,sh]
+-----------
+aws_sign_request:
```
untergeek: This should be `False` or `True` here, don't you think?

docs/asciidoc/configuration.asciidoc
```diff
@@ -420,6 +426,37 @@ IMPORTANT: You must set your <<hosts,hosts>> to the proper hostname _with_ port.
...
+-----------
+
```
untergeek: There should be a line somewhere indicating the default value (which I presume is `False`, right?).

curator/utils.py
```diff
@@ -745,36 +751,60 @@ def get_client(**kwargs):
...
+            raise ImportError('Failed to import a module: %s' % e)
+        try:
+            session = session.Session()
+            credentials = session.get_credentials()
+            kwargs['aws_key'] = credentials.access_key
+            kwargs['aws_secret_key'] = credentials.secret_key
+            kwargs['aws_token'] = credentials.token
+        # If an attribute doesn't exist, we were not able to retrieve credentials as expected so we can't continue
+        except AttributeError:
+            logger.debug('Unable to locate AWS credentials')
```
untergeek: Hmmm.  I'll let this slide, since we're raising the `NoCredentialsError` anyway, but it seems like this should be `logger.error`, rather than `logger.debug`.

untergeek:
Thanks for your work on this.  Hopefully it results in some very happy AWS users.

PR #1086, opened 2017-10-16
Use cached stats result to lookup names

tschroeder-zendesk:
Code was doing a client lookup for each name making it very slow to determine which node to use.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -8,8 +8,9 @@ Changelog
 **Bug Fixes**
 
   * The ``restore_check`` function did not work properly with wildcard index
-    patterns.  This has been rectified, and an integration test added to 
+    patterns.  This has been rectified, and an integration test added to
     satisfy this.  Reported in #989 (untergeek)
+  * Re-use cached response for nodes in most_available_node function. #1086 (tschroeder-zendesk)
 
 5.1.1 (8 June 2017)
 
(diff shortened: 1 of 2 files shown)


curator/actions.py
```diff
@@ -1901,7 +1901,7 @@ def most_available_node(self):
         mvn_id = None
         nodes = self.client.nodes.stats()['nodes']
         for node_id in nodes:
-            name = self.client.nodes.stats()['nodes'][node_id]['name']
+            name = nodes[node_id]['name']
```
untergeek: 🤦‍♂️ Thanks for this.  Can't believe I didn't think of that.

untergeek: Please add a comment to the Changelog on this in the bug fixes section.

tschroeder-zendesk: Done!

Changelog.rst
```diff
@@ -8,34 +8,35 @@ Changelog
 **Bug Fixes**
 
   * The ``restore_check`` function did not work properly with wildcard index
-    patterns.  This has been rectified, and an integration test added to 
+    patterns.  This has been rectified, and an integration test added to
     satisfy this.  Reported in #989 (untergeek)
+  * Re-use cached response for nodes in most_available_node function
```
untergeek: Can you mention the PR # and your github handle in parenthesis, like the other lines?

curator/actions.py
untergeek: I think know why it's like that now.  I probably cut and pasted what I was using in `ipython` to test (it wasn't in a loop, there).

Changelog.rst
tschroeder-zendesk: 👍

PR #1106, opened 2017-11-22

cjuroz:


untergeek:
Thanks for this change.  Please merge it against the `5.x` branch instead of `master`.  I've already got a huge code change against master locally, and this will not work there.  I'd rather let you include it into a working branch than have to redact it and have you have to start over.  This approach makes _me_ do the work to re-incorporate your work into my new branch.

untergeek:
Unless this is changed to merge against `5.x` as base, I cannot merge this.

cjuroz:
I was on vacation these days :-) I created a new PR #1114 so I am going to close this one.

PR #1107, opened 2017-11-27
Remove unit_count parameter doc for parameter that no longer exists

dashford:
Was still showing up in the documentation even though the `unit_count` parameter has been removed.


untergeek:
Please change this to be committed against the `5.x` branch, rather than `master`.  Huge changes have already been made to my local code branch for `master`, and this will actually conflict there.

Thanks!

dashford:
Hey @untergeek, I've made those changes now - let me know if there's anything else needed here.

setup.py
```diff
@@ -42,7 +42,6 @@ def get_install_requires():
 
 
     base = 'Console'
-    msvcrt = ''
```
untergeek: Why is this causing issues for you?

untergeek: Nevermind.  That was one of the other commits from changing base.

PR #1111, opened 2017-11-29
Dockerfile that builds from source and produces slim image

mikn:
Tested and verified working against the 5.4 branch. Does however not work against master right now, because build_exe isn't working in master.
We publish our image here: https://quay.io/repository/lalamove/elasticsearch-curator

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:3.6-alpine3.6 as builder
 
 RUN apk --no-cache add build-base tar musl-utils openssl-dev
-RUN pip3 install setuptools cx_Freeze==6.0b1 requests-aws4auth
+RUN pip3 install setuptools cx_Freeze==6.0b1 requests-aws4auth boto3
 
 COPY . .
 RUN ln -s /lib/libc.musl-x86_64.so.1 ldd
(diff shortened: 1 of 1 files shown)


untergeek:
Please commit against the `5.x` branch, rather than 5.4, please.

mikn:
@untergeek Sorry, you're right! There you go. :)

Dockerfile
```diff
@@ -1,8 +1,16 @@
-# Docker Definition for ElasticSearch Curator
+FROM python:3.6-alpine3.6 as builder
 
-FROM python:2.7.8-slim
-MAINTAINER Christian R. Vozar <christian@rogueethic.com>
+RUN apk --no-cache add build-base tar musl-utils openssl-dev
+RUN pip3 install setuptools cx_Freeze==6.0b1 requests-aws4auth
```
untergeek: You might want to also add `boto3` and `botocore` if someone is going to be using AWS Elasticsearch.  See [here](https://github.com/elastic/curator/blob/master/curator/utils.py#L847-L848)

untergeek:
@christianvozar What do you think of these changes?  I ask, since it was your build originally.

untergeek:
@christianvozar, if you have no objections, I will likely go ahead with these changes.

christianvozar:
@untergeek This seems very reasonable; anything to get image sizes smaller while maintaining same functionality is more than welcome. 👍

PR #1114, opened 2017-12-01

cjuroz:

Note: PR opened against `5.x` per author request in PR #1106


untergeek:
Hmmm.   With these conflicts, you might be better off rebasing, or kill this PR and do another against the 5.x branch (check out `5.x` locally instead of `master`)

cjuroz:
Created PR #1116

PR #1116, opened 2017-12-02

cjuroz:

Note: PR opened against 5.x per author request in PR #1106


untergeek:
Thanks for merging against `5.x` (I know that was an unnecessary hassle).

PR #1117, opened 2017-12-05
Add missing `exclude: True` in `timestring` docs

GregMefford:
## Proposed Changes

  - Add missing `exclude: True` in `timestring` docs


untergeek:
Why do you feel like the `exclude` must be here?  I made it agnostic, so you could either include or exclude the data by adding that line.

untergeek:
Thanks for adding this.

PR #1118, opened 2017-12-05
Stop overriding verify_certs param in AWS ES

igalarzab:
Hi!,

I'm configuring curator against an AWS ES using a CNAME registry and it's failing when the certificate is being validated (as the domain name it's different). I can see in the code that curator doesn't respect the value of the `ssl_no_validate` config param in this specific case.

I'm deleting the override as I don't see any reason to force the value to be True

### DIFF
--- a/curator/utils.py
+++ b/curator/utils.py
@@ -866,6 +866,8 @@ def get_client(**kwargs):
             # Override these kwargs
             kwargs['use_ssl'] = True
             kwargs['verify_certs'] = True
+            if kwargs['ssl_no_validate']:
+                kwargs['verify_certs'] = False
             kwargs['connection_class'] = elasticsearch.RequestsHttpConnection
             kwargs['http_auth'] = (
                 AWS4Auth(


untergeek:
Rather than delete this line, perhaps set `validate_certs` to the value of `ssl_no_validate`?

untergeek:
Also, please use `5.x` as the base, not `master`, as there have been significant changes to master in my current branch, and these won't merge nicely anymore.

igalarzab:
Changed the base branch. About the value of `ssl_no_validate`, I assumed the default value was being assigned here:

https://github.com/igalarzab/curator/blob/0884f0d312d39fea54d413e321e6a503ed814147/curator/utils.py#L794

Correct me if I'm wrong

igalarzab:
Sorry, forget what I've said, I misunderstood your comment! Now I'm respecting the `ssl_no_validate` value :)

untergeek:
I got it wrong, though.  It should be more like [here](https://github.com/igalarzab/curator/blob/0884f0d312d39fea54d413e321e6a503ed814147/curator/utils.py#L806-L807), as `ssl_no_validate` == `True` is the same thing as `kwargs['verify_certs']` == `False`.

untergeek:
Also, have you tried the new `boto` authentication? See https://www.elastic.co/guide/en/elasticsearch/client/curator/5.4/configfile.html#aws_sign_request

igalarzab:
Yes, I'm actually using that authentication mechanism

curator/utils.py
```diff
@@ -865,7 +865,8 @@ def get_client(**kwargs):
         if kwargs['aws_key']:
             # Override these kwargs
             kwargs['use_ssl'] = True
-            kwargs['verify_certs'] = True
```
untergeek: You can probably just leave this line as it is, and then add the two new lines to override the `kwargs['verify_certs'] = True` if `ssl_no_validate` is `True`.  Otherwise, you'll need an `else` statement to assign it to `True`.

untergeek:
This looks good now.  Thanks for adding it.  Will merge after Travis CI is green.

curator/utils.py
igalarzab: 🤦‍♂️ Yes, completely right, sorry.

Actually, just out of curiosity, as they are booleans we probably can just use `kwargs['verify_certs'] = not kwargs['ssl_no_validate']`

igalarzab:
cool, and thanks for the great support! 😄

curator/utils.py
untergeek: True.  But I like the code readability (for new users) of a default being set, and then overridden.

This will all get cleaned up for version 6 anyway, though.

PR #1119, opened 2017-12-06
[refactor] improve exception management

4383:
Fixes #

## Proposed Changes

  - Improve exception management
  - Use docstring of module instead string parameter to print function


untergeek:
This is great.  Can you use `5.x` as the branch instead of `master`?  This will make it into 5.4.1 if you can do that soon.

4383:
@untergeek Done! :)

4383:
Same refactor on all run commands

PR #1132, opened 2018-01-16
Fix body too big when we get metadata, create multiple query, one per…

rewiko:
Fixes Avoid querying all the indices in on go, the payload is too big depending on the number of indeces and the elasticsearch configuration. (specially on AWS when you can't customize the payload parameter) - getting HTTP/1.1 413 FULL head when you have too much indices.

## Proposed Changes

  - Making a request per index to get information

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,14 @@ RUN python3 setup.py build_exe
 
 FROM alpine:3.6
 RUN apk --no-cache upgrade && apk --no-cache add ca-certificates
-COPY --from=builder build/exe.linux-x86_64-3.6 /curator/
+COPY --from=builder build/exe.linux-x86_64-3.6 /curator-bin/
+RUN apk --no-cache add python py-setuptools py-pip gcc libffi py-cffi python-dev libffi-dev py-openssl musl-dev linux-headers openssl-dev libssl1.0 && \
+    pip install boto3==1.4.8 && \
+    pip install requests-aws4auth==0.9 && \
+    pip install cryptography==2.1.3 && \
+    apk del py-pip gcc python-dev libffi-dev musl-dev linux-headers openssl-dev
+COPY curator-entrypoint /usr/bin/curator
+RUN chmod 755 /usr/bin/curator
+RUN chmod 755 /curator-bin/curator
 USER nobody:nobody
-ENTRYPOINT ["/curator/curator"]
+ENTRYPOINT ["/usr/bin/curator"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,14 +11,6 @@ RUN python3 setup.py build_exe
 
 FROM alpine:3.6
 RUN apk --no-cache upgrade && apk --no-cache add ca-certificates
-COPY --from=builder build/exe.linux-x86_64-3.6 /curator-bin/
-RUN apk --no-cache add python py-setuptools py-pip gcc libffi py-cffi python-dev libffi-dev py-openssl musl-dev linux-headers openssl-dev libssl1.0 && \
-    pip install boto3==1.4.8 && \
-    pip install requests-aws4auth==0.9 && \
-    pip install cryptography==2.1.3 && \
-    apk del py-pip gcc python-dev libffi-dev musl-dev linux-headers openssl-dev
-COPY curator-entrypoint /usr/bin/curator
-RUN chmod 755 /usr/bin/curator
-RUN chmod 755 /curator-bin/curator
+COPY --from=builder build/exe.linux-x86_64-3.6 /curator/
 USER nobody:nobody
-ENTRYPOINT ["/usr/bin/curator"]
+ENTRYPOINT ["/curator/curator"]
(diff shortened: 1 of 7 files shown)


untergeek:
This sounds like an AWS Elasticsearch-specific problem here.  Curator specifically got away from making a bajillion redundant API calls because that hammers and traumatizes your cluster.  I'm really not open to this, but I'm willing to listen to why you'd change behavior that's been 100% fine for Curator users since 4.x to suit a narrow use-case that is AWS Elasticsearch-specific.

rewiko:
I think for any cluster with a big number of indices, your payload will become bigger and bigger. At some point we will reach the limit for the payload size. I was thinking to do it in a smarter way, fetching a number x of indices per requests. Then we will have a few requests to get metadata.

untergeek:
@rewiko You mention:

> specially on AWS when you can't customize the payload parameter  - getting HTTP/1.1 413 FULL head

Out of a vested interest in pursuing or denying this PR, how many indices do you have?  How many nodes in the cluster?  How many shards per index?

rewiko:
@untergeek Actually the cluster is quite small, 7 nodes 32 gb, number of index is 420. I think It's more related to the number of index and the mappings associated to (not a big mapping definition), curator will get the metadata for all the indices in one request.
The parameter to modify is 
```
http.max_content_length | The max content of an HTTP request. Defaults to 100mb. If set to greater than Integer.MAX_VALUE, it will be reset to 100mb.
-- | --
``` 
source https://www.elastic.co/guide/en/elasticsearch/reference/5.5/modules-http.html

We are currently using the Instance type r3.2xlarge.elasticsearch where the http.max_content_length is 100 MB.

untergeek:
You've provided the node count and the index count.  However, I really want to know the number of shards per index.  The number of shards per node can have a dramatic effect on cluster performance, if it is too big.

What constitutes "not a big mapping definition?"  A 100MB cluster state metadata size is quite large, and [Curator only pulls the metadata](https://github.com/elastic/curator/blob/v5.4.1/curator/indexlist.py#L152-L153), not even the rest of the cluster state.

rewiko:
Number of shard is 4 per index and 0 replica.

example of mapping
```
 {
      "template": "logs-{{ env }}-*",
      "settings": {
        "number_of_shards": 4,
        "number_of_replicas": 0
      },
      "mappings": {
        "flb_type": {
          "dynamic": true,
          "dynamic_templates": [
            {
              "custom_app_fields" : {
                "match" : "*",
                "match_mapping_type" : "string",
                "mapping": { "type": "keyword" }
              }
            }
          ],
          "properties": {
            "message": {
              "type": "text"
            },
            "@shipper-timestamp": {
              "type": "date",
              "format": "strict_date_optional_time||epoch_millis"
            },
            "@timestamp": {
              "type": "date",
              "format": "strict_date_optional_time||epoch_millis"
            },
            "docker": {
              "properties": {
                "container_id": {
                  "type": "keyword"
                }
              }
            },
            "kubernetes": {
              "properties": {
                "annotations": {
                  "type": "object"
                },
                "container_name": {
                  "type": "keyword"
                },
                "host": {
                  "type": "keyword"
                },
                "labels": {
                  "properties": {
                    "app": {
                      "type": "keyword"
                    },
                    "deployment": {
                      "type": "keyword"
                    },
                    "k8s-app": {
                      "type": "keyword"
                    },
                    "kubernetes_io/cluster-service": {
                      "type": "keyword"
                    },
                    "pod-template-hash": {
                      "type": "keyword"
                    },
                    "version": {
                      "type": "keyword"
                    }
                  }
                },
                "cluster": {
                  "type": "keyword"
                },
                "namespace_name": {
                  "type": "keyword"
                },
                "pod_id": {
                  "type": "keyword"
                },
                "pod_name": {
                  "type": "keyword"
                }
              }
            },
            "log": {
              "type": "text"
            },
            "stream": {
              "type": "keyword"
            }
          }
        }
      }
    }
```

config curator action:

```
actions:
      1:
        action: delete_indices
        description: "Delete indices older than 21 of days"
        options:
          ignore_empty_list: True
          continue_if_exception: False
        filters:
            - filtertype: age
              source: name
              direction: older
              timestring: '%Y.%m.%d'
              unit: days
              unit_count: 21
```

that's means we query "_cluster/state/metadata/" with 400 index in on go, so it get bigger than 100MB.
A bulk option to fetch metadata per X (ex:10 by 10)makes more sense for me.

untergeek:
Actually, I think it makes more sense to catch a failure and process in batches if that fails.

rewiko:
@untergeek ok let me update the PR

untergeek:
Please catch that 413 specifically for this retry in batches, and fail on any other failure.

rewiko:
I can't reproduce locally the huge payload even with hundred of indices, let's close it for the moment.

rewiko:
I was able to reproduce the bug, I've updated the PR.

untergeek:
Out of curiosity, why did you rename a bunch of the methods in the class?

PR #1154, opened 2018-02-28
Fix to be compliant with AWS ES 5.1

wanix:
Fixes #880 

## Proposed Changes

  - change the source index settings if we don't have the awaited default source
  - tested with AWS ES versions 5.1, 5.3, 6.0

Issue #880, opened 2017-02-07
AWS Elasticsearch version 5.x support?


xamox:
AWS just released their Elasticsearch service last week (https://aws.amazon.com/about-aws/whats-new/2017/01/elasticsearch-5-now-available-on-amazon-elasticsearch-service/).   It seems like curator works with AWS but version 4 doesn't appear to work with AWS ES v2.x, will version 4 support AWS ES v5.x?

untergeek:
See the [version compatibility matrix](https://www.elastic.co/guide/en/elasticsearch/client/curator/current/version-compatibility.html).

As it appears that AWS ES now supports the /_cluster/state endpoint, it _should_ work with Curator.

xamox:
@untergeek Thanks for the quick feedback.  Okay, I'll give it a test run and see how it goes.

volcomism:
I run into the following issue when trying to connect to an ES cluster (v5.1.2) hosted by AWS. Running this locally against the same ES version, works fine. Has anyone else been able to get this working?

Version 4.2.6

Command: 
```
curator_cli --logfile ./curator.log --loglevel INFO --logformat default --host aws-elasticsearch-domain --port 80 delete_indices --filter_list '[{"filtertype":"age","source":"creation_date","direction":"older","unit":"days","unit_count":13},{"filtertype":"pattern","kind":"prefix","value":"logstash"}]'
```

Traceback:
```
Traceback (most recent call last):
  File "/usr/local/bin/curator_cli", line 9, in <module>
    load_entry_point('elasticsearch-curator==4.2.4', 'console_scripts', 'curator_cli')()
  File "/usr/local/lib/python2.7/dist-packages/curator/curator_cli.py", line 5, in main
    cli(obj={})
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
[...]

PR #1172, opened 2018-03-16
change the source index settings if we don't have the awaited default source

wanix:
cancel and replace https://github.com/elastic/curator/pull/1154

Fixes # https://github.com/elastic/curator/issues/880

## Proposed Changes
Compliance with AWS ES 5.x


untergeek:
Can you add a blurb to the Changelog to reflect these changes, please?

PR #1195, opened 2018-04-09
Update utils.py

jacoor:
update doc.

## Proposed Changes

  - documentation update


untergeek:
Sorry.  This is rather arbitrary, and unnecessary, as an Elasticsearch repository can only be identified by its name in the API anyway.

PR #1201, opened 2018-04-19
Update filters.asciidoc - Clarity on Allocated Filter

fatalglitch:
Add clarity on how this filter performs, as it is a bit reverse from the other filter defaults

Fixes #

## Proposed Changes

  - Add notes about how the default's behave, they are different than other filter types


untergeek:
Thanks for the patch, but I prefer the approach to this in #1203, so I'm going with that change instead.

PR #1208, opened 2018-05-03
#1206 - The unit_count_pattern setting can cause indices to mistakenly be included in an index filter

### LINKED ISSUE #1206: The unit_count_pattern setting can cause indices to mistakenly be included in an index filter
When defining an age filter with the setting unit_count_pattern and no fallback value the current implementation can cause indices to mistakenly be included or excluded in the filtered list.

Example action:

```
actions:
  1:
    action: delete_indices
    description: Delete indices older than the defined retention time from the index name
    options:
      ignore_empty_list: True
      timeout_override:
      continue_if_exception: False
      disable_action: False
    filters:
    - filtertype: age
      source: name
      direction: older
      timestring: '%Y.%m.%d'
      unit: days
      unit_count: -1
      unit_count_pattern: 'logstash-\w+-([0-9]+)-[0-9]{4}\.[0-9]{2}\.[0-9]{2}'
      exclude:
```

This would work fine for the following index list:
logstash-test-2-2018.04.25
logstash-test-2-2018.04.25
logstash-test-2-2018.04.25
logstash-test-2-2018.04.25

However if you add an index like this:
logstash-te-st-2-2018.04.25
this causes the regular expression to fail once and reverses the decision on all indices that are evaluated afterwards.

The issue comes from [this line of code](https://github.com/elastic/curator/blob/master/curator/indexlist.py#L468) which causes an index that did not match the defined pattern to be ignored, if no fallback value has been configured.
However, since indices are iterated over in a loop exclude stays true for all indices following the non-matching one, causing the exact opposite of the actual intended behavior. This can cause catastrophic data loss if used with a delete option.

I've prepared a fix, will do some additional testing and open a PR shortly.

soenkeliebau:
Fixes #1206 

Introduce and extra boolean to keep track of indices that failed to match the pattern without changing the value for _exclude_ for the rest of the loop execution.

Added a test case that forces the bug to happen with non-fixed code.

### DIFF
--- a/curator/indexlist.py
+++ b/curator/indexlist.py
@@ -502,6 +502,7 @@ class IndexList(object):
                 unit_count_matcher = None
         for index in self.working_list():
             try:
+                removeThisIndex = False
                 age = int(self.index_info[index]['age'][self.age_keyfield])
                 msg = (
                     'Index "{0}" age ({1}), direction: "{2}", point of '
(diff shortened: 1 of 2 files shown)


untergeek:
Thanks for the fix!

PR #1218, opened 2018-05-16
Filter lifecycle mgmt indices by default

### LINKED ISSUE #1207: Curator should ignore indices with index.lifecycle.name set
In an upcoming version of Elasticsearch, we expect to have built-in index lifecycle management (ILM) for several common cases.  In order to avoid ILM and Curator "fighting for" control of an index (e.g. allocating and reallocating to different nodes based on partially or fully duplicative policies), we should check for indices that ILM is "in control of" and ignore them.  Having `index.lifecycle.name` set is such an indicator.

untergeek:
Look for index setting `index.lifecycle.name`.  Unless the option `allow_ilm_indices` is true, filter them by default.

This was also added for all singletons.

fixes #1207


tomsommer:
Somehow I feel the default should have been "false"

PR #1222, opened 2018-05-21
Fix show_indices and show_snapshots singletons

untergeek:
These were not working as expected


untergeek:
Travis is having issues:
```
E: Failed to fetch http://mirror.jmu.edu/pub/ubuntu/dists/trusty-updates/universe/i18n/Translation-en  Writing more data than expected (243993 > 243863)
E: Some index files failed to download. They have been ignored, or old ones used instead.
The command "sudo apt-get update && sudo apt-get install oracle-java8-installer" exited with 100.
```
A sample of the actual tests shows they're passing.

PR #1229, opened 2018-05-29
add extra pip packages to requirements for es aws

willejs:
Fixes #

Errors at runtime when trying to use aws es due to missing deps.

## Proposed Changes

Add required requirements for AWS elasticsearch.
At the moment if you don't install these dependencies and are trying to sign aws requests, the code errors at runtime.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,5 @@ click>=6.7
 pyyaml>=3.10
 voluptuous>=0.9.3
 certifi>=2017.7.27.1
+boto3>=1.7.28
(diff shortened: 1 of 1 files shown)


untergeek:
[These are already fixed in the 5.4.4 release](https://github.com/elastic/curator/blob/v5.5.4/requirements.txt).

willejs:
@untergeek cheers. I didnt realise that master wasnt the latest code and tags wernt cut against it and instead there were branches for each version.

PR #1259, opened 2018-08-16
Change Connection Class

### LINKED ISSUE #510: Proxy support for Curator
It would be nice to be able to pass a proxy option to curator. This way I can have things network isolated like I want, and run Curator behind something I have to proxy into.

raynigon:
Fixes #510

## Proposed Changes

Changed the Connection Class.
This leads to a usage of the HTTP_PROXY and HTTPS_PROXY enviroment variables.
Since this is standard Elasticsearch client functionality, there shouldnt be any tests needed at this point.


raynigon:
@untergeek Is this a way to go, or do you have a better idea to solve the Proxy Problem?

untergeek:
> Is this a way to go, or do you have a better idea to solve the Proxy Problem?

Maybe, but I won’t be merging client connection code changes in this repository. I will be using 
https://github.com/untergeek/es_client here in the next major release, and that’s where a proxy solution would have to go to get in.  

It’s more involved than just changing the connection class to use Requests instead of `urllib3`, too, as that’s still the safer/saner default, where possible (at least in my mind).  Perhaps use `Requests` _instead_ of the default when a proxy is requested, rather than to wholly replace everything as it currently exists.  A considerable amount of testing would be have to be added to outright replace `urllib3` with `Requests, but perhaps it would work just fine.

untergeek:
So yes, to recap, I did completely spin out the client code into its own class and project at 
https://github.com/untergeek/es_client and plan on using that for Curator and other projects in the future.

raynigon:
When do you think this will be available for Curator?

untergeek:
Redirect to #1438

PR #1263, opened 2018-08-28
Added filtertype 'empty' to support filter empty indices

jrask:
## Proposed Changes
Added filtertype: empty to support filtering of empty indices.

The function was already created, I have only exposed it as a filtertype and added documentation.

### DIFF
--- a/curator/defaults/filtertypes.py
+++ b/curator/defaults/filtertypes.py
@@ -138,3 +138,8 @@ def state(action, config):
         filter_elements.state(),
         filter_elements.exclude(),
     ]
+
+def empty(action, config):
+    return [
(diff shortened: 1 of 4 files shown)


elasticcla:
Hi @jrask, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/curator/pull/1263.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

jrask:
Need to create a new PR with correct e-mail, not sure if I am allowed to use the one the used in this one.

PR #1264, opened 2018-08-29
Exposed filter_empty function as a filter

jrask:
## Proposed Changes

It is now possible to filter empty indices. Simply exposing the function that already existed.

```
- filtertype: empty
  exlude: True|False
```


jrask:
@untergeek  - Seems like the errors where not related to my fix but stalled build during installation of java. Is it possible to force a travis rebuild?

untergeek:
Yes.  I just restarted the two failed builds.

PR #1272, opened 2018-09-12
Update command-line.asciidoc

zafarella:
## Proposed Changes
- add reference to curator_cli

### DIFF
--- a/docs/asciidoc/command-line.asciidoc
+++ b/docs/asciidoc/command-line.asciidoc
@@ -246,3 +246,7 @@ Exit codes will indicate success or failure.
 * `-1` - Exception raised that does not result in a `1` exit code.
 
 &nbsp;
+
+
+See also:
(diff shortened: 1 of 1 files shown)


untergeek:
I'm sorry, I just don't see this as a necessary addition to this page.  The breadcrumb at the top refers to that page, and the sidebar does too.

zafarella:
true, just saw it, but frankly you cannot notice it until you actually search deeply.

I was looking into the way listing snapshots and this is how came to this page.

Mbe add it as "Info" or "Tip" section?

untergeek:
The exit codes refer to both the `curator` and `curator_cli` binaries.  It shouldn't need more than that, really.

zafarella:
I think this is what I meant https://github.com/elastic/curator/pull/1273
Sorry for confusing, not much familiar with ascidocs yet

PR #1277, opened 2018-09-20
Delete redundant paragraph

gallardo:
Minor fix
## Proposed Changes

  - Delete redundant paragraph


untergeek:
It is redundant on purpose.  I don’t expect all users to read through both sections, and I wanted that portion to be in each for that reason.

PR #1290, opened 2018-10-12
Sort index list alphabetically prior to sorting by age

tschroeder-zendesk:
## Proposed Changes

  - Updates initial sort of indices for age_sort to be alphabetical. This fixes an issue where you want to exclude or include from the top of a list that is only one day of results but could be more. Since all the dates are the same it does nothing for either reverse true or reverse false you just get the same semi-random list. By sorting alphabetically before the age sort you can sort reverse or not and have different results when the result set is in fact all from 1 day


untergeek:
Sorry for the delay.  Been on vacation.  Let me take a look at these changes and get back to you.

untergeek:
Just a question, how does this intersect with the [pattern feature in the `count` filter](
https://www.elastic.co/guide/en/elasticsearch/client/curator/5.5/filtertype_count.html#_pattern_based_sorting)? I ask because I see you using the `count` filter in your test, but the pattern feature seems to intersect with what you're trying to do.

tschroeder-zendesk:
Since the change is inside the _sort_by_age each pattern group would be sorted first alphabetically and then by age. So same affect on those indices. I can add an additional test on that if you would like.

untergeek:
Sorry for the delays.  I will hopefully get around to reviewing this further tonight.

PR #1291, opened 2018-10-21
Add checks for reindex task failures and number of docs created

msvechla:
Currently curator simply checks if the destination index was created after a reindex task.
This is not enough for running this in production for us, as we want to delete the source indices after a successful reindex.

Therefore it can be the case, that the reindex task finishes with failures and curator does not throw an exception. The following actions will still get executed and in our case this means the source indices get deleted, even though the reindex task completed with a lot of errors.

## Proposed Changes

  - check the reindex task status for failures
  - check if the number of created docs match the total docs
 
Would be awesome if you could take a look at this. The implementation here is just a first draft.

### DIFF
--- a/curator/actions.py
+++ b/curator/actions.py
@@ -1280,7 +1280,7 @@ class Reindex(object):
             del reindex_args['slices']
         return reindex_args
 
-    def _post_run_quick_check(self, index_name):
+    def _post_run_quick_check(self, index_name, task_id):
         # Verify the destination index is there after the fact
         index_exists = self.client.indices.exists(index=index_name)
         alias_instead = self.client.indices.exists_alias(name=index_name)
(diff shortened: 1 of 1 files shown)


untergeek:
Sorry for the delays.  I will hopefully get around to reviewing this further tonight.

untergeek:
I will look this all over in a few days.  I'm super busy with work right now.

curator/actions.py
```diff
@@ -1303,6 +1303,36 @@ def _post_run_quick_check(self, index_name):
...
+            raise exceptions.FailedExecution(
+                'Reindex completed with {0} errors. '
+                'Check task: {1} for more information.'.format(len(failures), task_id)
+            )
+
+        # Verify that the number of docs in the source and destination index match
+        total_docs = response['task']['status']['total']
+        created_docs = response['task']['status']['created']
+
+        if not created_docs == total_docs:
```
untergeek: So, devil's advocate here.  What happens if someone does a reindex from query instead of a full index? Will these still be equal? What if someone reindexes from multiple indices to 1 index (e.g. 1 week's worth of daily indices into a single  weekly index)? Do these numbers match up? If the answer is not "yes," then I can't use the code—at least not as-is.

The draft code works just fine in the event of a 1:1 reindex. I could merge it if the test is only made made when a `verify_doc_count` flag is present (or something like it).  It would, of course, have to be vetted by conditional tests.  It wouldn't be valid, for example, if Curator was able to detect a many-to-one reindex (code which would have to be added).

untergeek:
Cross-reference #1299

curator/actions.py
rahst12: If you reindex going from a mapping file without nested documents to a mapping file with nested documents, the document count will be different.  I assume the doc count verification should be optional...

untergeek: Thanks for providing a concrete example, @rahst12

untergeek:
Sorry for the hideous delay on this. I'm barely catching up. Can you rebase this so I can merge it?

PR #1298, opened 2018-10-24
cushind/adding-shards-filtertype

### LINKED ISSUE #1071: [FEATURE] Allow index filtering based on number of shards
Currently there is no way to filter indices on number of shards with the new shrink api this would be particularly useful as you cannot shrink something to 1 shard that already has 1 shard. So, by adding this filter as an exclude that error can be eliminated without having to ignore errors.

cushind:
Fixes #1071

## Proposed Changes

  - Add shard filter type to filter indices based on their number of shards.
  - Shrink implicitly filters out all indices whose shard count is not greater than the target shrink 
     number of shards.
  - Can also explicitly filter out indices whose shard count does not meet the criteria. Range 
     facilitated by multiple shard filters.


untergeek:
I will look this all over in a few days.  I'm super busy with work right now.

untergeek:
This is a good first draft. I'll merge this now, and add some documentation later.

PR #1325, opened 2018-12-05
Add indices count + indices names to all actions

shaharmor:
## Proposed Changes
  - Add the count of indices + indices names to all actions (Or at least the relevant ones) so its easier to get that info at the beginning of the action execution

### DIFF
--- a/curator/actions.py
+++ b/curator/actions.py
@@ -233,7 +233,7 @@ class Allocation(object):
         )
         self.index_list.filter_closed()
         self.index_list.empty_list_check()
-
+        self.loggit.info('Updating {0} selected indices: {1}'.format(len(self.index_list.indices), self.index_list.indices))
         self.loggit.info('Updating index setting {0}'.format(self.body))
         try:
             index_lists = utils.chunk_index_list(self.index_list.indices)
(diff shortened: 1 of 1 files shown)


curator/actions.py
```diff
@@ -581,7 +581,7 @@ def do_action(self):
         self.index_list.filter_forceMerged(
             max_num_segments=self.max_num_segments)
         self.index_list.empty_list_check()
-        self.loggit.info('forceMerging selected indices')
+        self.loggit.info('forceMerging {0} selected indices: {1}'.format(len(self.index_list.indices), self.index_list.indices)
```
untergeek: This line is not properly terminated, so the CI tests will not even run.

curator/actions.py
```diff
@@ -1019,7 +1019,7 @@ def do_action(self):
         seconds between retries.
         """
         self.snapshot_list.empty_list_check()
-        self.loggit.info('Deleting selected snapshots')
+        self.loggit.info('Deleting {0} selected snapshots: {1}'.format(len(self.snapshot_list.snapshots), self.snapshot_list.snapshots)
```
untergeek: This line is also missing a closing `)`

curator/actions.py
untergeek: that is to say, it needs an additional `)`

shaharmor: Whoops, sorry. Edited on github 👎

PR #1333, opened 2018-12-27
Fix docs typos in filter_period

sloria:
## Proposed Changes

  - Fix typos that were causing incorrect rendering. Addresses all Sphinx warnings.


untergeek:
Thanks for catching these! Catching quote issues can be a strain on the eyes.

PR #1355, opened 2019-01-22
Use re.search() not re.match() in pattern filter so regex doesn't have to be a prefix

### LINKED ISSUE #1354: regex in pattern filter only matches against start of string
### Expected Behavior
https://www.elastic.co/guide/en/elasticsearch/client/curator/current/filtertype_pattern.html#_regex suggests I can use a regex pattern to match anywhere in a string - for example use the pattern "y" to match an index name "xyz".

### Actual Behavior
regex only ever matches when the pattern starts at the beginning of the index name: for example index "xyz" matches "x" but not "y".

### Detailed Description
This happens because https://github.com/elastic/curator/blob/5.6/curator/indexlist.py#L451 (same in master at time of writing) uses the Python match() call, not search().  match() only evaluates the regex from the start of the string.

### Workarounds / Impact
It's confusing that this isn't clearly documented as a limitation.

You can work around this in some cases by inserting ".*" or equivalent at the beginning of the regex.  Unfortunately, I want to be able to pull the regex from an environment variable, but that doesn't support being combined with another string.  Overall, I want to be able to match indices with name "<constant text>.${environment variable}.*", which is currently impossible.  I think I'm going to have to add a startup wrapper to modify the environment variable and add .* to it.

matthewdupre:
Fixes #1354 

Issue says it all really: currently regex pattern filters only match against the start of the index name.  The prefix / suffix patterns are fine because they get wrapped with ^<prefix>.*$ etc.

### DIFF
--- a/curator/indexlist.py
+++ b/curator/indexlist.py
@@ -448,7 +448,7 @@ class IndexList(object):
         pattern = re.compile(regex)
         for index in self.working_list():
             self.loggit.debug('Filter by regex: Index: {0}'.format(index))
-            match = pattern.match(index)
+            match = pattern.search(index)
             if match:
                 self.__excludify(True, exclude, index)
             else:
--- a/curator/snapshotlist.py
+++ b/curator/snapshotlist.py
@@ -233,7 +233,7 @@ class SnapshotList(object):
         self.empty_list_check()
         pattern = re.compile(regex)
         for snapshot in self.working_list():
-            match = pattern.match(snapshot)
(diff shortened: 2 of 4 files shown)


untergeek:
Please sign the CLA, and I can merge this after the CI tests come back.

It might be nice to add this same change to the Snapshot list class as well.

matthewdupre:
Thanks for the suggestion @untergeek - snapshot list updated.  I think this should be ready now.

PR #1369, opened 2019-03-22
Specify Loader= when calling yaml.load

teluka:
DeprecationWarning raised by utils.py.

That's because new version of pyyaml which was released on Mar 13, 2019 deprecates the usage of yaml.load(input): https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation.

```
/var/lib/elasticsearch/.local/lib/python2.7/site-packages/curator/utils.py:53: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  return yaml.load(read_file(path))
```

### DIFF
--- a/curator/utils.py
+++ b/curator/utils.py
@@ -50,7 +50,7 @@ def get_yaml(path):
     yaml.add_constructor('!single', single_constructor)
 
     try:
-        return yaml.load(read_file(path))
+        return yaml.load(read_file(path), Loader=yaml.FullLoader)
     except yaml.scanner.ScannerError as err:
         print('Unable to read/parse YAML file: {0}'.format(path))
         print(err)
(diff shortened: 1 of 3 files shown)


untergeek:
I appreciate the effort here. The CI tests will continue to fail, however, until all of the `yaml.load` lines in the tests are also replaced.

teluka:
@untergeek I've updated test but CI build still fails...

untergeek:
They will fail, because apparently—and according to my own testing—another change to PyYAML will have to be made before it will work properly, and 5.1.x or 5.2 is not out yet.  Even then, it might not work as desired. I may have to pin to 3.12 and change how variables are read in for Curator v6.

untergeek:
This is moot as PyYAML is now 6.0.0 and functional

Issue #1407, opened 2019-05-27
Please pin pyyaml  to 3.13 until the pyyaml 5.1 fixes are done for python 3.7 support


bryder:
pyyaml-3.12 will not build the c extensions when using python 3.7 so I can not build curator using python 3.7.

pyyaml-3.13 added support for python-3.7 but curator is pinned to 3.12 at the moment.

[reference to changes in pyyaml](https://pyyaml.org/wiki/PyYAML)

This relates to #1368 

### Expected Behavior

pyyaml-3.13 should be the requirement so curator can be used  with python 3.7

### Actual Behavior

I can not build curator using python 3.7 if the requirements are respected 

### Steps to Reproduce the Problem

I use an internal build system behind a firewall so can't easily show the steps here.

I suspect installing python 3.7 and building with pip will show the error when trying to build pyyaml-3.12 which emit the same errors I've been getting but unfortunately I can't do that from where I am.

### Specifications

  - Version: curator 5.7.6
  - Platform: ubuntu 16.04/18.04, Python 3.7.3

untergeek:
This is done. Curator 5.8 is currently pinned ti pyyaml=3.13

PR #1410, opened 2019-06-04
Feature/api key config

### LINKED ISSUE #1409: ReadonlyREST plugin for Elasticsearch requires X-Api-key header to be set
### Expected Behavior
ReadonlyREST plugin requires X-Api-key to be set to the correct value, it should be possible to set this.

### Actual Behavior
It is not possible to configure Curator to set the X-Api-key header when accessing Elasticsearch.

  1. Run Elasticsearch with ReadonlyREST enabled and API key required to access Elasticsearch
  1. Try to run Curator
  1. It fails, because Curator can't set the X-Api-key header with the required API key

## Context (Environment)
Example readonlyrest configuration (set in elasticsearch.yml:
```
readonlyrest:
  access_control_rules:
    - name: "api key"
      type: allow
      api_keys: ["foo"]
```

## Detailed Description
I propose adding an api_key setting in the config file.

vetler:
Fixes #1409 in 4.3 branch

## Proposed Changes

  - adds the `api_key` option to `config.yml`
  - if `api_key` is set, adds the `x-api-key` header to HTTP requests 

Also fixes failing environment variable support as in #1407

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 urllib3>=1.8.3
 elasticsearch>=2.4.0,<3.0.0
 click>=6.0
-pyyaml>=3.10
+pyyaml==3.13
 voluptuous>=0.9.3
 certifi>=2017.4.17
--- a/curator/utils.py
+++ b/curator/utils.py
@@ -592,6 +592,8 @@ def get_client(**kwargs):
             )
     try:
         client = elasticsearch.Elasticsearch(**kwargs)
+        if 'api_key' in kwargs:
+          client.transport.connection_pool.connection.headers.update({'x-api-key': kwargs['api_key']})
         # Verify the version is acceptable.
         check_version(client)
         # Verify "master_only" status, if applicable
--- a/curator/validators/config_file.py
+++ b/curator/validators/config_file.py
@@ -20,6 +20,7 @@ def config_client():
         Optional('timeout', default=30): All(
             Coerce(int), Range(min=1, max=86400)),
         Optional('master_only', default=False): Boolean(),
+        Optional('api_key', default=False): Any(None, str, unicode)
     }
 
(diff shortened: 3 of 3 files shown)


elasticcla:
Hi @vetler, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/curator/pull/1410.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

vetler:
@elasticcla Sorry, should be fixed now.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 urllib3>=1.8.3
 elasticsearch>=2.4.0,<3.0.0
 click>=6.0
-pyyaml>=3.10
```
untergeek: You may need to rebase against master. This set of requirements is wrong. _Very_ outdated. This may be a part of what's causing your errors in CI. Look at the current [requirements.txt](https://github.com/elastic/curator/blob/master/requirements.txt) from the master branch to see what I mean.

vetler: No, this is a PR for the 4.3 branch, which has these dependencies. Yes, it is intentional, we need this in the 4.3 branch.

untergeek: Then you might as well maintain your own branch, then. I'm no longer publishing releases to the 4.x branch of Curator, as the last one of those was released the 10th of April 2017.

Why are you pinned on such an old branch? Do you really have that old a version of Curator you're trying to support? The 5.x releases support Elasticsearch 5.x - 7.x, so I am imagining Elasticsearch 2.x here, which is also EOL and no longer supported by Elastic.

vetler: Ok, not surprising. Yes, I know it's ancient :)

PR #1427, opened 2019-07-11
Adjust the docs for the empty filtertype to mention impact on closed indices

heyitsmdr:
The `empty` filtertype for curator will match not only on empty indices but also indices that have been closed which were not empty at the time of closing.


untergeek:
This is not to be corrected with a documentation change, then, but to automatically add the `closed` filter to the code, so as to automatically exclude closed indices from consideration.

heyitsmdr:
That’s a fair point. I will close this PR and address the change in code.
Thanks!

untergeek:
There should be examples of this in other `actions`, and possibly filters.

heyitsmdr:
Thanks! I'll take a look and open a PR when ready.

PR #1430, opened 2019-07-12
Exclude closed indices from consideration when using the empty filter

heyitsmdr:
## Proposed Changes

- When using the [empty](https://www.elastic.co/guide/en/elasticsearch/client/curator/current/filtertype_empty.html) filtertype, automatically exclude closed indices from consideration. This will allow users to use the empty filtertype in a cluster with closed indices and not accidentally target them.

### DIFF
--- a/curator/indexlist.py
+++ b/curator/indexlist.py
@@ -727,6 +727,9 @@ class IndexList(object):
         """
         Filter indices with a document count of zero
 
+        Indices that are closed are automatically excluded from consideration
+        due to closed indices reporting a document count of zero.
+
         :arg exclude: If `exclude` is `True`, this filter will remove matching
             indices from `indices`. If `exclude` is `False`, then only matching
             indices will be kept in `indices`.
@@ -734,6 +737,7 @@ class IndexList(object):
         """
         self.loggit.debug('Filtering empty indices')
         self.empty_list_check()
+        self.filter_closed()
         for index in self.working_list():
             condition = self.index_info[index]['docs'] == 0
             self.loggit.debug('Index {0} doc count: {1}'.format(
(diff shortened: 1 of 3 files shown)


hab278:
👍

curator/indexlist.py
```diff
@@ -728,13 +728,17 @@ def filter_empty(self, exclude=True):
...
+        due to closed indices reporting a document count of zero.
+
         :arg exclude: If `exclude` is `True`, this filter will remove matching
             indices from `indices`. If `exclude` is `False`, then only matching
             indices will be kept in `indices`.
             Default is `True`
         """
         self.loggit.debug('Filtering empty indices')
         self.empty_list_check()
+        self.filter_closed()
```
untergeek: Oooh, this should precede the line above it. 

Say for instance that you only had closed indices before this filter was called. In the present ordering, you would do the `empty_list_check()` first, followed by removing any closed indices (`filter_closed()`). This would result in an empty list condition which could have been caught by `empty_list_check()`.

Please put the `self.filter_closed()` line above the `self.empty_list_check()` line and this code change should be good to go.

heyitsmdr: Ahh, that makes sense! I'll move that line up.

docs/Changelog.rst
```diff
@@ -18,6 +18,9 @@ Changelog
     This addresses #1363, and everyone else upgrading to Elasticsearch 7.x.
     Update documentation accordingly. (untergeek)
 
+  * Prevent the ``empty`` filtertype from incorrectly matching against closed
+    indices (heyitsmdr)
```
untergeek: Please add the reference to this PR: #1430 (the other lines all refer to the issue or PR number)

PR #1437, opened 2019-07-20
add skip_flush option for close action

### LINKED ISSUE #1412: Close action fails when index has unassigned shards
### Expected Behavior
Close action should be able to close indices with unassigned shards.

### Actual Behavior
Close action fails closing indices with unassigned shards.

  1. Spin up a cluster with at least 2 data nodes
  1. Create an index
  1. Terminate a data node containing at least one of its shards (make the index go red)
  1. Invoke Curator's close action on the index

### Specifications

  - Version: 5.4.1 (By looking at the code it seem to also affect the latest, 5.7.6 as of today, but haven't tested yet)
  - Platform: Docker on Ubuntu 18.04
  - Subsystem: Python 2.7

## Context (Environment)
We use Curator for periodically restoring prod data on another cluster, the flow consists of 3 actions - close -> restore -> open (in case the restore fails and the index is left closed). Normally it works well, once every now and then a node gets terminated and we're left with unassigned shards (not an issue, intentionally not using replicas here), as we're constantly restoring the data we expect the restore flow to work in this case as well but we find ourselves manually closing indices each time it happens.

## Detailed Description
The `_close` API has no issues with unassigned shards, the actual part that fails is the `_flush` [API call right before that](https://github.com/elastic/curator/blob/master/curator/actions.py#L307). Obviously, in most cases, one would want to flush before closing an index, but the current implementation renders Curator's ability to restore pretty useless where it's most needed (e.g. data loss).
2 suggestions on how to overcome this issue:
1. Make the `_flush` API call optional via a `--skip-flush` flag.
2. Check for unassigned shards for each index and avoid flushing those.

Anyway - the `_flush` API is not an integral part of the `_close` API so it's kinda weird to enforce it in Curator's close action IMO

psypuff:
Fixes #1412

## Proposed Changes

  - Added a `skip_flush` option for the `close` action, defaults to `False`.

### DIFF
--- a/curator/actions.py
+++ b/curator/actions.py
@@ -254,12 +254,14 @@ class Allocation(object):
             utils.report_failure(e)
 
 class Close(object):
-    def __init__(self, ilo, delete_aliases=False):
+    def __init__(self, ilo, delete_aliases=False, skip_flush=False):
         """
         :arg ilo: A :class:`curator.indexlist.IndexList` object
         :arg delete_aliases: If `True`, will delete any associated aliases
             before closing indices.
         :type delete_aliases: bool
+        :arg skip_flush: If `True`, will not flush indices before closing.
+        :type skip_flush: bool
         """
         utils.verify_index_list(ilo)
         #: Instance variable.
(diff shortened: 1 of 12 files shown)


psypuff:
Hey @untergeek,
I totally forgot to add myself to the contributors file so I'd appreciate if you'll do. 😅  
Also, I see you've merged some 5.8 prep code, what's the ETA on that? We'd love to upgrade once you publish that.

psypuff:
Thanks!

Issue #1439, opened 2019-08-01
Curator_cli: filter suffix pattern behaves like a regex pattern for matching indicies


Daniel314:
Hello,

When the following command was executed, the filter pattern matched every single index (and consequently deleted every index). Is this expected behavior? From the documentation, it appears that a suffix match should be a substring match, NOT a regex match.

> $ curator_cli delete_indices --ignore_empty_list --filter_list '{"filtertype":"pattern","kind":"suffix","value":"/?"}'

The suffix pattern above ("/?") should not have matched any indicies. 
All of my indicies had a name of logstash-%Y.%m.%d, .kibana_1, .kibana_task_manager, or .reporting-%Y.%m.%d

Here's the output for a dry run:

> $ curator_cli --dry-run delete_indices --ignore_empty_list --filter_list '{"filtertype":"pattern","kind":"suffix","value":"/?"}'
> 2019-08-01 13:05:47,922 INFO      DRY-RUN MODE.  No changes will be made.
> 2019-08-01 13:05:47,922 INFO      (CLOSED) indices may be shown that may not be acted on by action "delete_indices".
> 2019-08-01 13:05:47,922 INFO      DRY-RUN: delete_indices: .kibana_1 with arguments: {}
> 2019-08-01 13:05:47,922 INFO      DRY-RUN: delete_indices: .kibana_task_manager with arguments: {}
> 2019-08-01 13:05:47,922 INFO      DRY-RUN: delete_indices: logstash-2019.07.29 with arguments: {}
> 2019-08-01 13:05:47,922 INFO      DRY-RUN: delete_indices: logstash-2019.07.30 with arguments: {}
> 2019-08-01 13:05:47,922 INFO      DRY-RUN: delete_indices: logstash-2019.07.31 with arguments: {}
> 2019-08-01 13:05:47,923 INFO      DRY-RUN: delete_indices: logstash-2019.08.01 with arguments: {}
> 2019-08-01 13:05:47,923 INFO      DRY-RUN: delete_indices: logstash-2019.08.02 with arguments: {}
> 2019-08-01 13:05:47,923 INFO      "delete_indices" action completed.

This is on Ubuntu Linux (14.04.6 LTS), installed from the http://packages.elastic.co/curator/5/debian APT repository.
I was initially using version 5.7.5 of the elasticsearch-curator APT package.  Upgrading to version 5.7.6 did not change the behavior.

[...]

PR #1443, opened 2019-08-02
#1439: Modify prefix and suffix filter regex patterns to behave like substring matches, and abstract out duplicated code related to this change

Daniel314:
Hello,

This is my first pull request.  This pull request is for elastic/curator issue #1439.

The Curator documentation is not clear about 'prefix' and 'suffix' filters being regular-expression based
To me, the documentation seems to imply that prefix and suffix patterns are substring matches, so this pull request turns the prefix and suffix filters into substring matches.

This pull request makes the following changes:
   - Common pattern filter code in curator/snapshotlist.py and curator/indexlist.py has been coalesced and moved to curator/utils.py
   - The regex patterns in curator/defaults/settings.py have been moved to curator/utils.py (the regex_map() function has been removed)
   - The new code in curator/utils.py for the pattern prefix and suffix searches escapes regex characters so the regex pattern behaves like a substring search
   - Clarifies in docs/asciidoc/inc_kinds.asciidoc that 'prefix' and 'suffix' matches are substring matches

If these changes aren't satisfactory, please give me some ideas about changes that would be acceptable.

Regarding running 'python setup.py test':
  - In my development environment, I'm using ElasticSearch version 7.1.
    The official elactic/curator 'python setup.py test' XML summary output at the bottom reports quite a few failures (the total "Coverage" is 72%) -- I'm not sure what I need to adjust in my environment to get everything to pass...
    some of the errors are unicode vs non-unicode string mismatches; other errors are errors like "AssertionError: 0 != 5".
  - I've run 'python setup.py test' on the modifications in this pull request, and the test_... lines at the top of the output are identical to that of the official elactic/curator 'python setup.py test' output on my system.
    Said another way: The tests that pass for the official code still pass with my changes.
  - The XML failure summary for my repo doesn't match the XML failure summary for the official elactic/curator repo (in my development environment) for the four python files that I've modified, but the results are similar.

  -  As such, someone with a full working development environment should re-run 'python setup.py test' to validate my changes.

Thank you,

 Daniel

### DIFF
--- a/curator/defaults/settings.py
+++ b/curator/defaults/settings.py
@@ -12,15 +12,6 @@ def version_min():
 def config_file():
     return path.join(path.expanduser('~'), '.curator', 'curator.yml')
 
-# Default filter patterns (regular expressions)
-def regex_map():
-    return {
-        'timestring': r'^.*{0}.*$',
-        'regex': r'{0}',
-        'prefix': r'^{0}.*$',
-        'suffix': r'^.*{0}$',
-    }
-
 def date_regex():
     return {
         'Y' : '4',
--- a/curator/indexlist.py
+++ b/curator/indexlist.py
@@ -427,23 +427,7 @@ class IndexList(object):
             Default is `False`
         """
         self.loggit.debug('Filtering indices by regex')
-        if kind not in [ 'regex', 'prefix', 'suffix', 'timestring' ]:
-            raise ValueError('{0}: Invalid value for kind'.format(kind))
-
-        # Stop here if None or empty value, but zero is okay
-        if value == 0:
-            pass
-        elif not value:
-            raise ValueError(
-                '{0}: Invalid value for "value". '
-                'Cannot be "None" type, empty, or False'
-            )
-
-        if kind == 'timestring':
-            regex = settings.regex_map()[kind].format(utils.get_date_regex(value))
-        else:
-            regex = settings.regex_map()[kind].format(value)
-
+        regex = utils.build_regex(kind, value)
         self.empty_list_check()
         pattern = re.compile(regex)
         for index in self.working_list():
--- a/curator/snapshotlist.py
+++ b/curator/snapshotlist.py
@@ -218,23 +218,7 @@ class SnapshotList(object):
             matching snapshots will be kept in `snapshots`.
             Default is `False`
         """
-        if kind not in [ 'regex', 'prefix', 'suffix', 'timestring' ]:
-            raise ValueError('{0}: Invalid value for kind'.format(kind))
-
-        # Stop here if None or empty value, but zero is okay
-        if value == 0:
-            pass
-        elif not value:
-            raise ValueError(
-                '{0}: Invalid value for "value". '
-                'Cannot be "None" type, empty, or False'
-            )
-
-        if kind == 'timestring':
-            regex = settings.regex_map()[kind].format(utils.get_date_regex(value))
-        else:
-            regex = settings.regex_map()[kind].format(value)
-
+        regex = utils.build_regex(kind, value)
         self.empty_list_check()
         pattern = re.compile(regex)
         for snapshot in self.working_list():
--- a/curator/utils.py
+++ b/curator/utils.py
@@ -200,6 +200,37 @@ def get_date_regex(timestring):
     logger.debug("regex = {0}".format(regex))
     return regex
 
+# Default filter patterns (regular expressions)
+def build_regex(kind, value):
+    """
+    Return a regex pattern based on the kind and value.
+
+    :arg kind: Can be one of: ``suffix``, ``prefix``, ``regex``, or
+        ``timestring``. This option controls the kind of regex pattern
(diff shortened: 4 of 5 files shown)


elasticcla:
Hi @Daniel314, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/curator/pull/1443.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Daniel314:
I've gone through the Contributor Agreement process twice now, and I'm still seeing a CLA signature missing error.  What else am I missing besides hopping through the DocuSign form process?

Daniel314:
@elasticcla : What e-mail address do you see my commit coming from?  I only have one e-mail address associated with my Github account.

Daniel314:
Okay, I believe I've submitted the CLA with the "correct" e-mail address now.

untergeek:
> The official elactic/curator 'python setup.py test' XML summary output at the bottom reports quite a few failures (the total "Coverage" is 72%) -- I'm not sure what I need to adjust in my environment to get everything to pass...
some of the errors are unicode vs non-unicode string mismatches; other errors are errors like "AssertionError: 0 != 5".

See [here](https://github.com/elastic/curator/blob/v5.7.6/travis-run.sh#L49-L74) for some of the node-specific settings required for testing. There will still be 4 skipped tests when testing against a single cluster (remote reindex won't be tested, but skipped).  These are fully tested in CI.

Regarding the code changes, I like and appreciate what you have added, but this may break functionality for users who have been relying on `prefix` and `suffix` being regular expressions. Keeping with the principle of least surprise, I will have to release this change with a major version upgrade, with warnings/notifications to users that this is how it works.

untergeek:
As an example, here is a semi-sanitized version of my testing `elasticsearch.yml` file:

```
cluster.name: foo
node.name: bar
path.repo: ["/", "C:\\" ]
reindex.remote.whitelist: esclient:9200, localhost:9200, 10.11.12.*:9200, localhost:9201
node.ingest: false
xpack.monitoring.enabled: false
node.ml: false
xpack.security.enabled: false
xpack.watcher.enabled: false
```

untergeek:
Also,

### Coverage: 97%
Module | statements | missing | excluded | coverage
-- | -- | -- | -- | --
Total | 3270 | 112 | 0 | 97%
curator/__init__.py | 11 | 0 | 0 | 100%
curator/_version.py | 1 | 0 | 0 | 100%
curator/actions.py | 949 | 24 | 0 | 97%
curator/cli.py | 136 | 2 | 0 | 99%
curator/config_utils.py | 38 | 1 | 0 | 97%
curator/defaults/__init__.py | 0 | 0 | 0 | 100%
curator/defaults/client_defaults.py | 6 | 0 | 0 | 100%
curator/defaults/filter_elements.py | 83 | 0 | 0 | 100%
curator/defaults/filtertypes.py | 66 | 2 | 0 | 97%
curator/defaults/option_defaults.py | 158 | 0 | 0 | 100%
curator/defaults/settings.py | 33 | 0 | 0 | 100%
curator/exceptions.py | 13 | 0 | 0 | 100%
curator/indexlist.py | 510 | 28 | 0 | 95%
curator/logtools.py | 36 | 0 | 0 | 100%
curator/repomgrcli.py | 80 | 1 | 0 | 99%
curator/snapshotlist.py | 199 | 0 | 0 | 100%
curator/utils.py | 840 | 54 | 0 | 94%
curator/validators/__init__.py | 1 | 0 | 0 | 100%
curator/validators/actions.py | 20 | 0 | 0 | 100%
curator/validators/config_file.py | 4 | 0 | 0 | 100%
curator/validators/filters.py | 32 | 0 | 0 | 100%
curator/validators/options.py | 13 | 0 | 0 | 100%
curator/validators/schemacheck.py | 41 | 0 | 0 | 100%

bryder:
This would be a breaking change which will break all of my curator configs. 

Ideally the docs would say they are regex's not substrings. 

I was also confused by the docs - I was working through someone else's curator config and was surprised to see regex's in the prefix patterns. 

I found the issue for the PR and discovered they are regexes.

untergeek:
> Ideally the docs would say they are regex's not substrings.

This was addressed in #1558 via some asciidoc notes, e.g.

> NOTE: Internally _regex_ pattern is constructed from prefix value as `^{0}.*$`. Special characters should be escaped with backslash to match literally.
> NOTE: Internally _regex_ pattern is constructed from suffix value as `^.*{0}$`. Specials character should be escaped with backslash to match literally.

You can see that it's merged into the 8.0 docs [here](https://github.com/elastic/curator/blob/8.0/docs/asciidoc/inc_kinds.asciidoc). It will be a few hours before it shows up in the main documentation.

I agree with @bryder here. If switching to substring matching is going to break things for people, then I can't make a breaking change like that. Informing people that it is regex based should suffice. There's simply too much history here to arbitrarily change it now.

PR #1453, opened 2019-08-14
example should have host without port affixed

dkm199:
Fixes # example for case when port will be affixed

## Proposed Changes

  - use localhost as example for host without post affixed instead of localhost:9202


docs/asciidoc/configuration.asciidoc
```diff
@@ -273,7 +273,7 @@ port: 9200
 -----------
 
 The default is `9200`.  This value will only be applied to <<hosts,hosts>>
-without a port affixed, e.g. `localhost:9202`.
+without a port affixed, e.g. `localhost`.
```
untergeek: If you really want to change this, I request that you change it instead to show a before and after, e.g.

"This value will only be applied to hosts _without_ a port affixed. For example, if you have:

```
  hosts: "localhost:9200"
  port: 9202
```
In this example, port `9202` will not be affixed to `localhost:9200`, nor will `9202` replace the `9200`."

docs/asciidoc/configuration.asciidoc
```diff
@@ -272,8 +272,11 @@ This should be a single value:
 port: 9200
 -----------
 
-The default is `9200`.  This value will only be applied to <<hosts,hosts>>
-without a port affixed, e.g. `localhost:9202`.
+The default is `9200`. This value will only be applied to hosts without a port affixed. For example, if you have:
+
+  hosts: "localhost:9200"
+  port: 9202
+In this example, port 9202 will not be affixed to localhost:9200, nor will 9202 replace the 9200.
```
untergeek: Please add a blank line between `port: 9202` and `In this example`. Asciidoc might be a bit unhappy otherwise.

untergeek:
This is no longer correct. I am closing it, as the 8.0 release now requires full URL schema with port.

Issue #1479, opened 2019-10-29
Add Freeze action to docs


sastorsl:
## Add Freeze action to the documentation for Curator.
Freeze action has been added to Curator in PR #1454 / issue #1399 
It is however not documented as pr https://www.elastic.co/guide/en/elasticsearch/client/curator/current/actions.html

### Expected Behavior
The new action to freeze should be documented.

### Actual Behavior
Docs are missing.

### Steps to Reproduce the Problem
N/A

### Specifications
  - Version: 5.8.1

## Context (Environment)
N/A

## Detailed Description
N/A

trondhindenes:
wow I had no idea. This definitely needs to go in the docs.

PR #1497, opened 2019-12-21
[DOC] Freeze and unfreeze documentation

lucabelluccini:
Fixes https://github.com/elastic/curator/issues/1479

I am not 100% sure if we support the options (it should as they are the "default" additional options).

Let me know if I should squash the commits.

## Proposed Changes

  - Add/fix examples for freeze/unfreeze
  - Add documentation on the action page for freeze/unfreeze


arizonagroovejet:
Is there any chance of this getting merged? I'm interesting in using Curator to freeze indices, I know from looking at issues that Curator has that functionality now. It would be nice if the documentation told people the functionality exists and how to use it.

PR #1500, opened 2020-01-14
Specify which github repo and project to get the package from

AndyHunt66:
## Proposed Changes
https://www.elastic.co/guide/en/elasticsearch/client/curator/5.8/python-source.html
The instructions for installing from source don't specify which github repository or project to download the tarball from - there's just a generic `username/project` in the url.

Changed to show the actual username (`elastic`) and the project (`curator`).

### DIFF
--- a/docs/asciidoc/installation.asciidoc
+++ b/docs/asciidoc/installation.asciidoc
@@ -303,9 +303,9 @@ That might look like this:
 
 [source,sh]
 --------------------------------------
-wget https://github.com/username/project/archive/#.#.#.tar.gz -O package.tar.gz
-tar zxf package.tar.gz
-cd package-#.#.#
+wget https://github.com/elastic/curator/archive/v#.#.#.tar.gz -O curator.tar.gz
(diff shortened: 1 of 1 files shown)


AndyHunt66:
Ah - ok - understood.

What's the main reason for expecting them to fork the project?
Is it simply because if they're not using the package-install mechanisms they'll probably want to be tweaking it anyway?

untergeek:
> Is it simply because if they're not using the package-install mechanisms they'll probably want to be tweaking it anyway?

Exactly. Chances are very good if you're doing this you've rolled your own and need to deploy it this way.

untergeek:
Any chance you'll update this to show both the "fork" method and what you've added? In other words, instead of replacing what you did, adding your bits after that instead, and include some details describing pulling the default repo vs. forking?

untergeek:
Thanks for this. This has been mitigated in the 8.0 release. There is only a single method documented for downloading the source package and installing via pip, with the actual file being properly (and dynamically) linked.

PR #1501, opened 2020-01-14
There was a “the” hanging in no mans land

geogdog:
I don’t think I need to explain this that much. To reproduce it you need to read the original sentence, and the read the correct one.


untergeek:
Thanks for catching this

geogdog:
You're totally welcome.

PR #1509, opened 2020-02-04
Update examples with index templates

pgomulka:
date math in ES is using java time where Y changed to y
this is in ES version 7.0 and newer


untergeek:
Please rebase against the current master branch. I'll look into anything that comes up after that. I know it's documentation only, but builds will fail unless you rebase.

PR #1510, opened 2020-02-04
Update date math section to use y instead of Y

pgomulka:
elasticsearch has changed to use y instead of Y as it migrated to use java.time
this is in ES version 7.0 and newer

### DIFF
--- a/docs/asciidoc/inc_datemath.asciidoc
+++ b/docs/asciidoc/inc_datemath.asciidoc
@@ -11,7 +11,7 @@ A date math name takes the following form:
 |===
 |static_name| is the static text part of the name
 |date_math_expr| is a dynamic date math expression that computes the date dynamically
-|date_format|is the optional format in which the computed date should be rendered. Defaults to `YYYY.MM.dd`.
+|date_format|is the optional format in which the computed date should be rendered. Defaults to `yyyy.MM.dd`.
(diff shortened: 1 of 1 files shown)


untergeek:
While this is true for Elastic 7.x, anyone using the older version can (and maybe should) keep using `Y`. Perhaps instructions or a `NOTE` or `WARNING` to that effect should be added?

pgomulka:
@untergeek 
if someone was using an older version, then indeed they can still use Y. However they can also use y.
before 70 the difference is very subtle
Y is year of era
y is just year
however since 7.0
Y is week-based-year
y is for year-of-era
u is for just year
I guess for this page it makes sense to have a warning, however it would be unclear for page  with examples https://github.com/elastic/curator/pull/1509/files
I wonder if it wouldn't be more clear to just keep the latest patterns on both pages. I have seen users raising questions on patterns and referring to curator documentation
WDYT?

untergeek:
It would be okay to change to `y` for all examples, though the `DD` might also need changing on those other examples, if you're going to that trouble.

An informative `NOTE` to explain that versions older than 7.0 use Joda time which uses different symbols for notation is worth adding.

pgomulka:
agree, will add notes

untergeek:
Please rebase against the current master branch. I'll look into anything that comes up after that.

PR #1522, opened 2020-03-05
Curator 5.6 is not compatible with ES 7.2

### LINKED ISSUE #1478: Compatibility matrix is slightly confusing
The compatibility matrix here: https://www.elastic.co/guide/en/elasticsearch/client/curator/5.x/version-compatibility.html

Shows that Curator 5 is compatible with ES 7. However, a customer running Curator 5.6 found that it was not compatible with ES 7.2.

Obviously this is because ES 7.2 (or indeed any 7.x) didn't exist when Curator 5.6 was released. However, the Matrix seems to suggest that ANY Curator 5 will work with ES 7.x. Now, if I change the docs version to Curator 5.6, then the ES 7.x column disappears, so it's correct. But a Google search tends to take you to the latest doc, so I can understand how the customer made the mistake.

I propose, therefore, a slight change to the rows of the table, to create a delineation between 5.6- and 5.7+. Or something similar.

jibsonline:
As mentioned in #1478

> Shows that Curator 5 is compatible with ES 7. However, a customer running Curator 5.6 found that it was not compatible with ES 7.2.
> 
> Obviously this is because ES 7.2 (or indeed any 7.x) didn't exist when Curator 5.6 was released. However, the Matrix seems to suggest that ANY Curator 5 will work with ES 7.x. Now, if I change the docs version to Curator 5.6, then the ES 7.x column disappears, so it's correct. But a Google search tends to take you to the latest doc, so I can understand how the customer made the mistake.

Fixes #1478

Split the matrix for 5 into two sections --> 5.6- and 5.7+ and updated the compatibility for ES7.x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,9 @@ python:
   - "3.5"
   - "3.6"
   - "3.7"
+  - "3.8"
 
 env:
-  - ES_VERSION=5.0.2
-  - ES_VERSION=5.1.2
-  - ES_VERSION=5.2.2
-  - ES_VERSION=5.3.3
-  - ES_VERSION=5.4.3
-  - ES_VERSION=5.5.3
   - ES_VERSION=5.6.16
   - ES_VERSION=6.0.1
   - ES_VERSION=6.1.3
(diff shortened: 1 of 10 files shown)


jibsonline:
I am not sure why the test would fail for a document update. Is there an issue with the pipeline?

untergeek:
Please rebase against the current master branch. I'll look into anything that comes up after that.

jibsonline:
@untergeek Done

Issue #1494, opened 2019-12-09
Shrink method hangs on ES 7.x AWS managed service


sbalaram:
## For usage questions and help
Please create a topic at https://discuss.elastic.co/c/elasticsearch
https://discuss.elastic.co/t/using-py-curator-shrink-method-hangs-on-es-7-x-aws-managed-service/210791

Perhaps a topic there already has an answer for you!

## To submit a bug or report an issue

shrink using curator python never completes and the cluster status goes to unassigned shards.

AWS support explanation :

Cause of the issue: From ES 7.0 , Elastic deprecated the use of "copy_settings" parameter in the shrink api call and made "copy_settings=true" as the default value. There is no way to modify this parameter anymore. Reference --> https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#copy-settings-deprecated-shrink-split-apis "copy_settings" --> allows users to specify whether they want the target index (shrunken index) to have the same index settings as the source index. Now, with "copy_settings" set to true in the shrink api calls for 7.x ES clusters, all the index settings from the target index including "index.routing.allocation.require._name" and "index.blocks.read_only" are also copied down to the target index. The above setting forces ES to assign shards of target index to a specific node that is defined in "index.routing.allocation.require._name" but with replicas enabled, ES cannot allocate replica of a primary shard onto the same datanode. So, Cluster settings prevent ES from allocating replicas onto the same datanode as primaries and index settings prevent the shards to be allocated to any other node but "index.routing.allocation.require._name" value. Previously, these settings were never inherited down to the shrunken indices as they

the workaround is to

Step1. Update the source index routing setting to a specific node and block all writes.
Step2. Perform Shrink operation.
Step3. Remove routing and write block settings from the shrunken index.

the workaround works, but I loose the benefit of using curator shrink DETERMINISTIC option ( https://www.elastic.co/guide/en/elasticsearch/client/curator/current/option_shrink_node.html ), automatically determines the node with most free space

anyone successfully ran shrink using curator on 7.x ES AWS cluster without any workarounds ?

### Expected Behavior

single call to shrink method worked on 6.x versions of ES, but the same fails on ES 7.x
### Actual Behavior

the method hangs and the cluster state changes to unassigned shards 

### Steps to Reproduce the Problem

  1. call shrink with DETERMINISTIC option and it the operation never completes 
  1.
  1.

### Specifications

  - Version: 7.x
  - Platform:AWS managed service
  - Subsystem:

[...]

PR #1528, opened 2020-03-10
Fix the hang of the shring method with ES 7.x

jclegras:
Fixes #

## Proposed Changes

Related to issue #1494.

With this fix, when we're shrinking we're doing as shown by the ES [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.6/indices-shrink-index.html#_shrinking_an_index) : 

```
POST /my_source_index/_shrink/my_target_index
{
  "settings": {
    "index.routing.allocation.require._name": null, 
    "index.blocks.write": null 
  }
}
```
Hence, the shrink index operation doesn't hang anymore because of the routing requirement.

It seems that all tests be played against a single ES instance.
This use case requires a cluster in order to reproduce this error with integration tests (need help).

### DIFF
--- a/curator/actions.py
+++ b/curator/actions.py
@@ -2095,8 +2095,8 @@ class Shrink(object):
             'settings': {
                 'index.number_of_shards' : number_of_shards,
                 'index.number_of_replicas' : number_of_replicas,
-                "index.routing.allocation.require._name": None,
-                "index.blocks.write": None
+                'index.routing.allocation.require._name': None,
+                'index.blocks.write': None
             }
         }
         if extra_settings:
--- a/curator/actions.py
+++ b/curator/actions.py
@@ -2091,16 +2091,23 @@ class Shrink(object):
         self.number_of_shards = number_of_shards
         self.wait_for_active_shards = wait_for_active_shards
         self.shrink_node_name = None
+
         self.body = {
             'settings': {
                 'index.number_of_shards' : number_of_shards,
                 'index.number_of_replicas' : number_of_replicas,
-                'index.routing.allocation.require._name': None,
-                'index.blocks.write': None
             }
         }
+
         if extra_settings:
             self._merge_extra_settings(extra_settings)
+
+        if utils.get_version(self.client) > (6, 0, 1):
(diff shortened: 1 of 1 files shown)


untergeek:
Tests are failing against ES 6.0.1:

```
INFO curator.actions.shrink              do_action:2437 Shrinking index "my_index" to "my_index-shrunken" with settings: {'settings': {'index.blocks.write': None, 'index.routing.allocation.require._name': None, 'index.number_of_replicas': 0, 'index.number_of_shards': 1}}, wait_for_active_shards=1
ERROR            curator.cli                    run:215  Failed to complete action: shrink.  <class 'curator.exceptions.FailedExecution'>: Exception encountered.  Rerun with loglevel DEBUG and/or check Elasticsearch logs for more information. Exception: Unable to shrink index "my_index" -- Error: TransportError(500, u'null_pointer_exception', None)
```
It may be that the `require` option does not work in ES 6.0 (or setting `index.blocks.write` to `None` does something unusual, though why I cannot fathom). I haven't dug deep enough to confirm that. If so, a version test will be required in the code to not use that option for ES < 6.1, and to use the previous code in ES 6.0.x.

jclegras:
All right, I will fix it, thx.

curator/actions.py
```diff
@@ -2091,17 +2091,24 @@ def __init__(
...
                 'index.number_of_replicas' : number_of_replicas,
-                'index.routing.allocation.require._name': None,
-                'index.blocks.write': None
             }
         }
+
         if extra_settings:
             self._merge_extra_settings(extra_settings)
 
+        if utils.get_version(self.client) > (6, 0, 1):
```
untergeek: Please make this `>= (6, 1, 0)` as all of the 6.0.x versions will have the API difference.

untergeek:
Thank you, the fix works, but please change the version check as indicated.

curator/actions.py
jclegras: Done

jclegras:
You're welcome, thank you for this tool.

PR #1529, opened 2020-03-13
Add ECS log format

### LINKED ISSUE #1519: ECS log format
Curator currently support JSON format but not in the Elastic Common Schema (ECS) format. I've added it to my own fork and wonder if that's a PR you would like to see here?

m1keil:
Add ECS log format to curator, which is just JSON but with field names defined in the Elastic Common Schema.

Fixes #1519

### DIFF
--- a/curator/cli_singletons/utils.py
+++ b/curator/cli_singletons/utils.py
@@ -155,7 +155,7 @@ def config_override(ctx, config_dict):
         elif k == 'host':
             if 'host' in ctx.params and ctx.params['host'] is not None:
                 config_dict['client']['hosts'] = ctx.params[k]
-        elif k in ['loglevel', 'logfile', 'logformat']:
+        elif k in ['loglevel', 'logfile', 'logformat', 'ecs']:
             if k in ctx.params and ctx.params[k] is not None:
                 config_dict['logging'][k] = ctx.params[k]
         else:
--- a/curator/defaults/client_defaults.py
+++ b/curator/defaults/client_defaults.py
@@ -38,6 +38,6 @@ def config_logging():
                 ),
         Optional('logfile', default=None): Any(None, *string_types),
         Optional('logformat', default='default'):
-            Any(None, All(Any(*string_types), Any('default', 'json', 'logstash'))),
+            Any(None, All(Any(*string_types), Any('default', 'json', 'logstash', 'ecs'))),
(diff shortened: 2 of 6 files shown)


untergeek:
A little late for the question, but, have you verified that using dotted field notation works properly? Or does it map the JSON as field names with dots in them? It's only in `master`, so we can update this if it does map strangely. I've seen it work sometimes, and not others, so I want to be sure.

untergeek:
Yeah. This won't work:

```
{"@timestamp": "2020-03-13T17:50:42.912Z", "log.level": "INFO", "log.logger": "curator.cli", "log.origin.file.line": 226, "log.origin.function": "run", "message": "Job completed."}
```

m1keil:
Both ECS and ElasticSearch supports dot notation and unpack the fields into objects, but I can do nested dicts if you prefer?

untergeek:
ECS and Elasticsearch support dot notation _for queries._ 

The output I shared shows that the dotted notation is inside the JSON as a key name. This will definitely be inserted/parsed by Elasticsearch as a key name of `{"foo.bar": "value"}` and not `{"foo": { "bar": "value"}}`

Not to worry, I have already re-coded the original `LogstashFormatter` class to address this. I'll share the PR number here when I submit it in a few minutes.

untergeek:
See #1531

untergeek:
```
{"@timestamp": "2020-03-13T22:39:38.041Z", "log": {"level": "INFO", "logger": "curator.cli", "origin": {"file": {"line": 226}, "function": "run"}, "original": "Job completed."}, "message": "Job completed."}
```

m1keil:
> ECS and Elasticsearch support dot notation _for queries._

Interesting. from my tests I assumed it just silently converts it [to objects as described here](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html).

However it seems like dot notation is still [causing some grief](https://github.com/elastic/elasticsearch/issues/51037) so I guess outputting proper objects is easier.

Thanks for the fix!

untergeek:
Yes. In 2.0, dotted notation was removed, then re-added in 2.4, then tweaked before being re-added for querying. Very confusing stuff. You see, if I'm querying something with dotted notation, I should eventually be able to find it. But if I'm trying to write with dotted notation, and one field in the path contains a period in the field name, then how is ES to know whether it should be nested or not?

So, for index writing, explicitly specifying nested paths is the best way to go.

PR #1530, opened 2020-03-13
Ensure proper API call is used

untergeek:
In older releases (< 7.5) of ES Python module, `client.tasks.get()` will return a list of all tasks. However, starting in 7.5, this results in a failed API call. See https://github.com/elastic/elasticsearch-py/issues/1110

Correct behavior is to use `client.tasks.list()` when no task id is provided.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 
 python:
   - "2.7"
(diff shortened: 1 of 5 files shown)


sethmlarson:
:+1:

PR #1531, opened 2020-03-13
Address nested object keys in ECS logging

untergeek:
Update Changelog.rst

### DIFF
--- a/curator/exceptions.py
+++ b/curator/exceptions.py
@@ -64,3 +64,8 @@ class ClientException(CuratorException):
     """
     Exception raised when the Elasticsearch client and/or connection is the source of the problem.
     """
+
+class LoggingException(CuratorException):
+    """
+    Exception raised when Curator cannot either log or configure logging
+    """
--- a/curator/logtools.py
+++ b/curator/logtools.py
@@ -3,38 +3,79 @@ import sys
 import json
 import logging
 import time
+from curator.exceptions import LoggingException
 
+def de_dot(dot_string, msg):
+    """Turn message and dotted string into a nested dictionary"""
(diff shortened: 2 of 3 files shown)


m1keil:
I would recommend using the `message` field instead of `log.original` in this case. 
> For log events the message field contains the log message, optimized for viewing in a log viewer.

While the `log.original`:
> This is the original log message and contains the full log message before splitting it up in multiple parts.
> 
> In contrast to the message field which can contain an extracted part of the log message, this field contains the original, full log message. It can have already some modifications applied like encoding or new lines removed to clean up the log message

Basically, you want to have `message` field, as it includes a human readable info about the event.

untergeek:
Yes, the `WANTED_ATTRS` for `ECSFormatter` maps `message` to `log.original`. But if you follow [this code block](https://github.com/elastic/curator/pull/1531/files#diff-9b34ba2d4d832a313abb6c65555cda1fR62-R66) you'll find that if `message` is not in the `result` dictionary, it puts it there no matter what.

There is good reasoning to including both. `log.original` is not analyzed, and is only stored in `_source` (as you correctly indicated). But it is used by the Kibana Logs UI.

`message` is just the default message, and in ECS is mapped as `text` and `keyword`, both.

untergeek:
Sample output:
```
{"@timestamp": "2020-03-13T22:39:38.041Z", "log": {"level": "INFO", "logger": "curator.cli", "origin": {"file": {"line": 226}, "function": "run"}, "original": "Job completed."}, "message": "Job completed."}
```
Note the presence of both `message` and `log.original` containing the same bits.

m1keil:
gotcha, thanks for clarifying that!

PR #1571, opened 2020-11-25
[DOCS] Update restore snapshot index setting links

jrodewig:
Updates several links to the [Changing index settings during restore](https://www.elastic.co/guide/en/elasticsearch/reference/master/snapshots-restore-snapshot.html#change-index-settings-during-restore) section of the ES reference. These work due to a redirect, but this will improve the UX.


jrodewig:
@untergeek No rush, but let me know if the CI failure is anything to be concerned about. I'm fairly sure that the link changes are correct.

PR #1574, opened 2020-11-30
Update pyyaml to 5.1

jmlrt:
This commit updates pyyaml version from 3.13 to 5.1 to fix
CVE-2017-18342[0].

[0]: https://github.com/advisories/GHSA-rprw-h62v-c2w7

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,6 @@ requests>=2.20.0
 boto3>=1.9.142
 requests_aws4auth>=0.9
 click>=6.7,<7.0
-pyyaml==3.13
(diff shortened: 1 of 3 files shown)


untergeek:
I appreciate the fix, but it's not that simple. You see, this represents a breaking change to the way environment variables can be used. It has to change for the newer version of pyyaml to work. That's why the CI tests are failing.  I am obliged to completely re-do how environment variables work, and that breaking change is necessary, but will undoubtedly upset a few people.

jmlrt:
Yep, closing it so.

PR #1575, opened 2020-11-30
omit get_list_object() and do_filters() for reindex operation with RE…

henkepa:
…INDEX_SELECTION

Fixes #

- skipping the unnecessary steps
self.get_list_object() & self.do_filters() when the index name is explicitly passed for a reindex operation.

Helpful in a scenario where self.get_list_object() would fail with 'index_not_found_exception'


Henkhogan:
Can this be merged, please?

Issue #1589, opened 2021-02-22
Dependency pinning urllib and click


vibpe:
### Expected Behavior
The most current dependencies should be used.

### Actual Behavior
In setup.py the dependency of urrlib3 is pinned to a version  <1.26 and the version of click is pinned to <7.0.

However curator 5.8.3 compiles here and works with urllib3 1.26.2 and click 7.1.2. Tested with Python 3.8 and Python 3.9.

What is the intention on pinning this dependencies? Are there any plans on upgrading this dependencies?

vibpe:
Gentoo bug regarding old dependencies of curator: https://bugs.gentoo.org/714860

untergeek:
The urllib dependency is because of upstream co-dependencies (Namely: boto3) that were broken at the time of last release. Click testing was similarly non-functional in the past. I can revisit these dependencies if `boto3` is finally properly functional with the latest `urllib3` releases.

sethmlarson:
@untergeek iirc boto3 now allows urllib3 1.26.x so you should be good to unpin.

untergeek:
These have been unpinned as of 5.8.4

PR #1595, opened 2021-03-23
unpin urllib and click

shtouff:
Fixes https://github.com/elastic/curator/issues/1589

## Proposed Changes

 The current pinning of deps makes curator unusable alongside the latest version of celery in a project for example. This PR removes that constraint.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,11 +23,11 @@ def get_version():
 
 def get_install_requires():
     res = ['elasticsearch==7.1.0' ]
-    res.append('urllib3>=1.24.2,<1.26')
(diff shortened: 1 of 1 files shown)


untergeek:
So, there's much more to this than merely unpinning. To make urllib3 work requires re-pinning the version of the `elasticsearch` module to the latest as well.

shtouff:
> So, there's much more to this than merely unpinning. To make urllib3 work requires re-pinning the version of the `elasticsearch` module to the latest as well.

yup, sorry for the noise, I wanted to run the CI before proposing it to review.

PR #1604, opened 2021-05-31
add .dockerignore to increase build speed

harduino:
On development when you change Dockerfile and run building Docker image, it starts from scratch, because with command "COPY . ." Dockerfile is copied too which forces Docker Builder to ignore building cache. We can prevent that by adding .dockerignore file with added line "Dockerfile". In future more files could be added to .dockerignore to increase build speed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+Dockerfile


untergeek:
Thank you for your contribution. As with #1605, however, I cannot merge it without the CLA being signed.

PR #1605, opened 2021-05-31
fix segfault

harduino:
On build latest version I found strange bug - if copy build-folder to any folder with other name - curator shows error "segmentation fault". I guess it's because build-script defines hard-coded path in compiled binary and as curator's folder moved to /curator/, its binary can't import compiled libraries from old path.
In this PR we add curator's binary to PATH, but leaved it in build-folder, so curator will run without "Segmentation fault" error.


untergeek:
Thank you for your contribution. I cannot merge it without the CLA being signed, however.

harduino:
Looks like this PR fix segmentation error not in all cases and the error still appears.

Issue #1606, opened 2021-06-04
'api_key' parameter is not processed correctly


ssavrim:
### Expected Behavior
```
2021-06-04 09:47:48,385 INFO      Creating client object and testing connection
2021-06-04 09:47:48,398 INFO      Instantiating client object
2021-06-04 09:47:48,399 INFO      Testing client connectivity
2021-06-04 09:47:48,903 INFO      Successfully created Elasticsearch client object with provided settings
```

### Actual Behavior
```
2021-06-04 09:41:58,123 INFO      Preparing Action ID: 1, "alias"
2021-06-04 09:41:58,124 INFO      Creating client object and testing connection
2021-06-04 09:41:58,129 INFO      Instantiating client object
2021-06-04 09:41:58,130 INFO      Testing client connectivity
2021-06-04 09:41:58,611 ERROR     HTTP 401 error: action [cluster:monitor/main] requires authentication
2021-06-04 09:41:58,612 CRITICAL  Curator cannot proceed. Exiting.
```
### Steps to Reproduce the Problem
Here the configuration I uses to execute curator

```
  client:
    hosts: my_elasticsearch_cloud_instance
    use_ssl: True
    certificate: '/etc/ssl/certs/ca-certificates.crt'
    ssl_no_validate: False
    api_key: my_base64_api_key
    timeout: 30
    master_only: False
```

### Specifications
  - Version: curator 5.8.3
  - Platform: linux

## Context (Environment)
We need to push execute curator on cluster which can be accessed only with 'api_key'.
Ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html

## Detailed Description
I succeeded to connect to our cluster by commenting following line in the source code.
https://github.com/elastic/curator/blob/3ddfbb95ea7a164bb3f8c773b0392481f067da7c/curator/utils.py#L1013

Normally elasticsearch.client natively supports 'api_key' field
https://elasticsearch-py.readthedocs.io/en/v7.13.1/index.html?highlight=api_key#api-key-authentication

untergeek:
Moot. This is addressed in 8.0 and via the newer `es_client` python module.

PR #1411, opened 2019-06-04
New configuration parameter: api_key - used in x-api-key header for r…

vetler:
Fixes #1409

## Proposed Changes

  - adds the `api_key` option to `config.yml`
  - if `api_key` is set, adds the `x-api-key` header to HTTP requests

PR #1607, opened 2021-06-04
Add api_key authentication support

ssavrim:
## Proposed Changes

  - add new method process_apikey_auth_args method
  - update documentation
  - add --apikey_auth option to run_singleton.py

#1606


untergeek:
So, straight up, will this push break the functionality added in #1411 ? You've refactored that function which allows an external plugin's api-key functionality to work. I'm concerned that it will break functionality that has been there for 2 years, and may be relied upon by others.

ssavrim:
Indeed I did not have all the context, sorry for that.
I will make another proposal to keep backward compatibility.

ssavrim:
@untergeek finally I implemented a new method to process api_key field.
This will avoid to break existing feature.

PR #1612, opened 2021-07-22
Add filter by size

### LINKED ISSUE #1151: [FEATURE] Add per index_space based filtertype
I' set up a hot-warm elastic cluster. The Indices in the cluster have different sizes (from 50mb to 200GB) with different shard count (3 up to 10).

Once i moved the index to the 'warm' nodes, it will be useful if i can shrink it to a number of shards based on it's size (usually related to the node heap memory). 
For example, we can shrink a <4GB index with 3 shards to 1 shard. Or perhaps indexing between 4GB and 8GB to 2 shards and so on. It has less impact moving  4GB shard size rather than 200GB shard size in case of a shard reallocation or node failure. 

Instead of changing the 'shrink' action to permit a shard number based on index size, maybe it will be more useful to just implement an index_space filtertype. This way we can apply different actions (shrink, reallocate, ...) based on the index size.

Thanks in advance

IndraGunawan:
Fixes  #1151

## Proposed Changes

  - Add size filter type to filter indices based on their their primary only size or total size.

### DIFF
--- a/curator/actions.py
+++ b/curator/actions.py
@@ -148,7 +148,7 @@ class Alias(object):
                 '"{3}"'.format(
                     job.rstrip('e'),
                     index,
-                    'to' if job is 'add' else 'from',
+                    'to' if job == 'add' else 'from',
                     alias
                 )
             )
--- a/curator/defaults/filter_elements.py
+++ b/curator/defaults/filter_elements.py
@@ -151,7 +151,7 @@ def timestring(**kwargs):
         return {Optional('timestring', default=None): Any(None, *string_types)}
 
 def threshold_behavior(**kwargs):
-    # This setting is only used with the space filtertype and defaults to 'greater_than'.
+    # This setting is only used with the space and size filtertype and defaults to 'greater_than'.
     return {
         Optional('threshold_behavior', default='greater_than'): Any('greater_than', 'less_than')
     }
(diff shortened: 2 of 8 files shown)


untergeek:
Thank you for the pull request. I'm curious. This is an unusual feature, at least to me. What's the use case? Why filter by size? I can see some value in an edge case where I have a ton of tiny indices and I have re-indexed everything to a much larger index and want to purge the tiny indices. But outside of an edge case, I (me personally) can't see where this is useful. That doesn't mean I won't merge this PR, I'm just curious.

untergeek:
Clever! I get it now! That's a clever way to address your need.

So to follow up in my curiosity, why not use rollover indices and just let them grow to a size? Is there a compelling reason to keep absolute daily indices rather than just letting them fill shards until they're between 20GB and 40GB?

IndraGunawan:
It's related to data retention. For example we want to keep last 1 month in the `hot tiers`, then after that `1 month + 1 day` indices we do a daily snapshot lets say `production-logs-20210622` and put it to a repository and we have snapshot retention that will remove 1 year old snapshot.

Someday we want to restore snapshot (or specific indices) on specific date lets say `production-logs-20210501`. By keeping the daily indices we can be sure that the data will be there because when we do snapshot we filter that specific date indices. i don't know how to do it if we do rollover on the indices, or i missed something?

By adding this capability we can reduce number of shards in `hot tiers` storage.

IndraGunawan:
@untergeek any updates on this?

untergeek:
Apologies. I had to be away from work for an unexpected funeral and an expected wedding. I should hopefully be able to address this shortly.

Issue #1611, opened 2021-06-23
Replacing non-inclusive term "blacklist" with neutral language


mrudrego:
Curator has a configuration property logging.**blacklist**
There are concerns raised on usage of this non-inclusive terminology. Linux team has approved on replacing these non-inclusive terminologies. There are many other tech companies and open-source projects that have removed references to racially-charged jargon from their code for more neutral and inclusive language. Reference [https://www.zdnet.com/article/linux-team-approves-new-terminology-bans-terms-like-blacklist-and-slave/]

This request is to replace the term blacklist used in one of the curator configuration with a neutral word say denylist or blocklist.

Thanks,

PR #1623, opened 2021-10-09
Replacing non-inclusive term blacklist with neutral language

chetanugale:
Fixes https://github.com/elastic/curator/issues/1611

### DIFF
--- a/curator/config_utils.py
+++ b/curator/config_utils.py
@@ -4,7 +4,7 @@ from copy import deepcopy
 from voluptuous import Schema
 from curator.validators import SchemaCheck, config_file
 from curator.utils import ensure_list, get_yaml, prune_nones, test_client_options
-from curator.logtools import LogInfo, Whitelist, Blacklist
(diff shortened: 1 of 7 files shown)


untergeek:
Stale. To have this PR considered, the following conditions must be met:

- Re-open (prefer re-create) against current branches (must also include 6.x and 7.x)
- Will not accept a PR that results in existing configurations failing. In other words, `blacklist` must remain acceptable and not break or prevent an existing user's script from functioning after an upgrade. I appreciate what you're doing here, but breaking existing functionality is not acceptable. We can log a deprecation message if `blacklist` is used, but its presence must not break functionality and usability. We can fully remove in a major version release, but not until it's been deprecated and logged as such for quite a while.
- Will not accept a PR that doesn't address documentation in both the rST/Sphinx ReadTheDocs documentation as well as the asciidoc docs that render on elastic.co). Change like this must be adequately documented. Cannot just drop this on users without explanation, warning, documentation and examples.

PR #1624, opened 2021-10-12
Use assertEqual instead of assertEquals for Python 3.11 compatibility.

tirkarthi:
## Proposed Changes

  Use assertEqual instead of assertEquals for Python 3.11 compatibility. Deprecated unittest aliases were removed in Python 3.11 . This will cause error when tests are executed in Python 3.11 .

Ref : python/cpython#28268


untergeek:
This is appreciated, but no longer necessary.

PR #1636, opened 2022-01-28
Resync with upstream 20220106

danpie-cw:
Fixes #resynced with Elasticsearch 5.8.4 to get up to date with most recent elasticsearch and support python 3.8

## Proposed Changes

  -
  -
  -


danpie-cw:
Just going to use elasticsearch-curator instead of our custom elasticsearch-curator-api to reduce maintenance.

PR #1651, opened 2022-08-29
Doc fix

jeanfabrice:
Fix  Python Boolean case

## Proposed Changes
  - Replace `true` with `True`
  - Replace `false` with `False`
 
at various places in the options documentation

### DIFF
--- a/docs/asciidoc/options.asciidoc
+++ b/docs/asciidoc/options.asciidoc
@@ -105,14 +105,14 @@ IMPORTANT: Read more about Curator and Index Lifecycle Management
 action: delete_indices
 description: "Delete the specified indices"
 options:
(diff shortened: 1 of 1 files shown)


untergeek:
I apologize for the inconvenience, but the `6.0` and `7.x` branches (and the as-yet un-pushed `8.x` branch) are where all of the future actions are going to be. These are much more likely to be merged if you do PRs against those two branches, rather than `master`.

untergeek:
Also recognize that this is a `YAML` file you're trying to force `Python` boolean conventions on. The `PyYAML` import translates it appropriately.

jeanfabrice:
Closing as irrelevant (see last comment from Aaron above)

PR #1652, opened 2022-11-12
Use Python's build in unittest.mock

jelly:
Since Python 3.3 the unittest module supports mock and replaces python-mock which is now deprecated.

Fixes #

## Proposed Changes

  - Stop using the deprecated mock module and use the Python unittest.mock module

### DIFF
--- a/curator/utils.py
+++ b/curator/utils.py
@@ -134,7 +134,7 @@ def verify_client_object(test):
     :rtype: None
     """
     # Ignore mock type for testing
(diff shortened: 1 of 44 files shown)


untergeek:
I appreciate the effort, but you should check out the 6.x and 7.x branches of Curator. I haven't updated master as I am working on a very new, very different 8.x branch at the same time.

jelly:
Ok, I'll close this PR then.

PR #1702, opened 2024-02-21
fix(#1701): Bump pyyaml requirement to 6.0.1 for PEP-517.

### LINKED ISSUE #1701: bug: Support PEP-517 in 6.x
## Support PEP-517 in 6.x

The 6.x release series presently doesn't support PEP-517 build systems (so things like poetry totally fail to install if `poetry add elasticsearch-curator@^6` for example.

### Expected Behavior

Running `pip wheel --no-cache-dir --use-pep517 "elasticsearch-curator (==6.0.0)"` should succeed.

### Actual Behavior

Fails due to `pyyaml==6.0.0` as opposed to `pyyaml==6.0.1`.

  1. pip wheel --no-cache-dir --use-pep517 "elasticsearch-curator (==6.0.0)"
  1. poetry add elasticsearch-curator@^6

### Specifications

  - Version: 6.0.0
  - Platform: Linux/MacOS/Windows/Anything
  - Subsystem: N/A

## Context (Environment)

We are presently instrumenting a lambda function to handle curation for a 6.x ES cluster. This isn't possible using contemporary build tools as a result of the explicit pinning to `6.0.0` over the `6.0.1` patch release addressing PEP-517.

## Detailed Description

See above. I can provide further details if required, but I am raising a PR for this against `6.x`.

mbeacom:
Fixes #1701 

## Proposed Changes

  - Update `pyyaml` package pin from `6.0.0` to `6.0.1` (which supports PEP-517 for modern build systems) specifically for `6.x` releases.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
     "boto3>=1.24.54",
     "requests_aws4auth>=1.1.2",
     "click==8.1.3",
-    "pyyaml==6.0.0",
+    "pyyaml==6.0.1",
     "voluptuous>=0.13.1",
     "certifi>=2022.12.7",
     "six>=1.16.0",


mbeacom:
@untergeek I'm uncertain if you're the appropriate contact, but is there anything you may need to push this quick change out or should I plan to publish a patched variant to accommodate for the incompatible dependency?
Thanks!

PR #1703, opened 2024-02-24
drop six dependency

a-detiste:
I don't know what `Any()` does, maybe code can be further simplified

https://wiki.debian.org/Python3-six-removal


a-detiste:
@cla-checker-service : ping

untergeek:
Thank you for this, but I ended up removing the need for `string_types` altogether via #1716

PR #1712, opened 2024-04-12
Fix unescaped characater error in docstring

idanceinbetween:
There is an unescaped character in the docstring of `indexlist` module which has been [deprecated since Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior).

This causes an issue during `pytest`'s test collection phase when running tests on modules that import curator as a library on later versions of Python: `"SyntaxError: invalid escape sequence '\d'"`

In popular python linters e.g. ruff and flake8 this [violates rule W605](https://docs.astral.sh/ruff/rules/invalid-escape-sequence/).

Upon linting the entire source code, this is the only file that has this rule violation. This commit fixes this error.
LintForW605

The change has been tested by modifying a local copy of the library and running my test suite that originally failed against it, and the exception no longer occurs.

## Proposed Changes
* Fix docstring unescaped character error, specifically for 7.x releases.

### DIFF
--- a/curator/indexlist.py
+++ b/curator/indexlist.py
@@ -883,7 +883,7 @@ class IndexList(object):
         source='creation_date', timestring=None, field=None,
         stats_result='min_value', exclude=True):
         # pylint: disable=W1401
-        """
+        r"""
         Remove indices from the actionable list beyond the number `count`,
         sorted reverse-alphabetically by default.  If you set `reverse` to
         `False`, it will be sorted alphabetically.


untergeek:
Thank you for this. I just patched this in the master, 8.x, and 8.0 branches as well. I have a lot of tech debt with the 7.x branch that is just hanging there.

idanceinbetween:
> Thank you for this. I just patched this in the master, 8.x, and 8.0 branches as well. I have a lot of tech debt with the 7.x branch that is just hanging there.

Ah yes, I have just seen it on the later branches. Unfortunately we cannot yet upgrade to 8.0 so would be fab to have at least this patched for 7.x too. Is there anything else that would be useful to get this approved, merged and released?

jiaxuanhon:
hi @untergeek, would you be able to approve this? Thanks!

idanceinbetween:
Hi @untergeek , can this PR be approved please?

untergeek (approved):
LGTM™

untergeek:
Sorry for the long delay. I've been super busy.

jiaxuanhon:
No worries - thank you for this!

idanceinbetween:
No worries - thank you for this!

PR #1728, opened 2024-09-13
Fix/name based ages

### LINKED ISSUE #1727: name-based age filter is not filtering indices properly
We recently updated our Elastic cluster from `7.17` to `8.15.1` and we also moved from Curator `5.8.4` to `8.0.16`. 

After some run, we spotted some issues that we didn't detect before. 
Some indices that do not match provided timestring are deleted.

Let me add some examples to be more clear 

# Use case 1 
Index : `index-prod`  (no timestring at all)
curator config : 
```yaml
---

actions:
   monthly:
   action: delete_indices
   description: 'Delete old monthly indices'
   options:
      continue_if_exception: false
      ignore_empty_list: true
   filters:
      - filtertype: age
        source: name
        direction: older
        timestring: '%Y.%m'
        unit: months
        unit_count: 2
```

Expected result : `index-prod` should not be touched since it does not match the timestring.
Actual result : `index-prod` is deleted

# Use case 2
Lack of documentation reading explained this usecase, as stated [here](https://github.com/elastic/curator/issues/1727#issuecomment-2350142509). 
<details>
<summary>
I just kept it collapsed so following discussion does not become strange with missing references.
</summary>

Index : `index-prod-2024.09`  (timestring with year and month)
curator config : 
```yaml
---

actions:
   monthly:
   action: delete_indices
   description: 'Delete old daily indices'
   options:
      continue_if_exception: false
      ignore_empty_list: true
   filters:
      - filtertype: age
        source: name
        direction: older
        timestring: '%Y.%m.%d'
        unit: days
        unit_count: 5
```

Expected result : `index-prod-2024.09` should not be touched since it does not match the timestring.
Actual result : `index-prod-2024.09` is deleted

</details>

# Logs showing the issue 
[curator.log](https://github.com/user-attachments/files/17492745/curator.log)

# Findings
After digging a little bit, I think I found the culprit (but not really sure about potential side effects).
When retrieving an index, `indexlist` [populates](https://github.com/elastic/curator/blob/265ac487daaf6e3db07c174c0f81e3a41737b341/curator/indexlist.py#L256) `self.index_info` with [zero_values](https://github.com/elastic/curator/blob/265ac487daaf6e3db07c174c0f81e3a41737b341/curator/indexlist.py#L116)

However, in `_get_name_based_ages`, we handled the timestring as : 
- Look for the timestring and get epoch
- If something has been found and returned an int value, [let's populate the age/name attribute](https://github.com/elastic/curator/blob/265ac487daaf6e3db07c174c0f81e3a41737b341/curator/indexlist.py#L427)

But what is happening if the epoch is not an int (e.g. None, if the epoch can't be provided since the timestring does not match anything) ? 
Well, since the age/name property has been initialized to 0, it stays as is.

And since 0 < point of reference , it gets deleted.

Just for the explanation, I understood that it was working with version `5.8.4` because index settings initialization is quite different and `age` property [was initialized with `{}`](https://github.com/elastic/curator/blob/71bd33463147d34d9252872e986f349cb7442ad9/curator/indexlist.py#L82)

# Workaround

For the moment, to workaround this, we have to add an additional filter that will match the pattern, like
```yaml
       - filtertype: pattern
         kind: regex
         value: ^index-prod-\d{4}\.\d{2}\.\$
```
for the second test.

# Potential fix
I'll propose a fix in a PR (with 2 additional tests). 
The main idea is to discard any index that do not match the timestring by setting the age/name to `sys.maxsize`

However I'm not sure if it is the best fix or if something smarter could be done.

giom-l:
Fixes #1727 

## Proposed Changes

Discard any index that do not match the timestring provided to avoid keeping the 0 age value.
Add 2 tests that were failing before this fix.

### DIFF
--- a/curator/indexlist.py
+++ b/curator/indexlist.py
@@ -2,6 +2,8 @@
 import re
 import itertools
 import logging
+import sys
+
 from elasticsearch8.exceptions import NotFoundError, TransportError
 from es_client.helpers.schemacheck import SchemaCheck
 from es_client.helpers.utils import ensure_list
@@ -426,6 +428,8 @@ class IndexList:
             epoch = tstr.get_epoch(index)
             if isinstance(epoch, int):
                 self.index_info[index]['age']['name'] = epoch
+            else:
+                self.index_info[index]['age']['name'] = sys.maxsize
 
 
     def _get_field_stats_dates(self, field='@timestamp'):
--- a/tests/integration/test_delete_indices.py
+++ b/tests/integration/test_delete_indices.py
@@ -55,6 +55,7 @@ class TestActionFileDeleteIndices(CuratorTestCase):
                 'age', 'name', 'older', '\'%Y.%m.%d\'', 'days', 30, '_([0-9]+)_', ' ', ' ', ' '))
         self.invoke_runner()
         self.assertEqual(15, len(exclude_ilm_history(get_indices(self.client))))
+
     def test_retention_from_name_days_keep_exclude_false_after_failed_match(self):
         # Test extraction of unit_count from index name and confirm correct
         # behavior after a failed regex match with no fallback time - see gh issue 1206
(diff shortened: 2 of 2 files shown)


curator/indexlist.py
```diff
@@ -426,6 +428,8 @@ def _get_name_based_ages(self, timestring):
             epoch = tstr.get_epoch(index)
             if isinstance(epoch, int):
                 self.index_info[index]['age']['name'] = epoch
+            else:
+                self.index_info[index]['age']['name'] = sys.maxsize
```
untergeek, line 432: This is a clever way to guarantee it will never be a 0 😂

untergeek:
Thank you for this. I really appreciate the effort put into diagnosing the problem, and submitting the Github Issue.

In the end, I chose to remove the non-matching indices rather than just put a very large number in the epoch time spot (which is a clever work-around).

PR #1737, opened 2025-02-26
[docs] Migrate docs from AsciiDoc to Markdown

colleenmcginnis:
Migrate docs from AsciiDoc to Markdown. The preview can be built after #1738 is merged.

@lcawl  please tag reviewers, add the appropriate label(s), and take this out of draft when you're ready.

Notes:
* I had to remove `docs/` from `.gitignore` to commit these changes.
* The `docs/` directory also contains a bunch of `rst` files. I didn't touch those files.

PR #1740, opened 2025-02-28
[docs] Clean up cross-repo links

colleenmcginnis:
Follow up to https://github.com/elastic/curator/pull/1737

When using cross-repo links, the path should be relative to the docset.yml not the full path within the repo ([updated docs-builder docs](https://elastic.github.io/docs-builder/syntax/links/#cross-repository-links)).

### DIFF
--- a/docs/docset.yml
+++ b/docs/docset.yml
@@ -1,6 +1,7 @@
 project: 'Curator index management'
 cross_links:
   - beats
+  - docs-content
   - elasticsearch
   - logstash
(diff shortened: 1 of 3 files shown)


untergeek:
These cross-repo links are so cool! I love Markdown already, so being able to do this in Markdown is going to be so refreshing, even if the doc builder is what fleshes it all out.

untergeek (approved):
LGTM™

PR #1763, opened 2025-05-15
[docs] Add `products` to `docset.yml`

colleenmcginnis:
Related to https://github.com/elastic/docs-builder/issues/1200

Add `products` to `docset.yml` to be used in the search experience.

cc @KOTungseth

### DIFF
--- a/docs/docset.yml
+++ b/docs/docset.yml
@@ -1,4 +1,6 @@
 project: 'Curator index management'
+products:
+  - elasticsearch
 cross_links:
   - beats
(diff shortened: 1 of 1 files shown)


docs/docset.yml
```diff
@@ -1,4 +1,6 @@
 project: 'Curator index management'
+products:
+  - elasticsearch
```
colleenmcginnis, line 3:
```suggestion
  - id: elasticsearch
```

PR #1796, opened 2026-04-28
Update option_name.md

erikfinnman:
Add 'rollover' to the list of actions used by the 'name' option.


erikfinnman:
I have signed the contributor agreement.

